Issue #2, opened 2018-12-29
Example server crashes on login abort


Reeywhaar:
Examples server's `/private_data` request crashes if oauth login was interrupted.

**Steps to reproduce**
* start example
* navigate to http://127.0.0.1:8080/web
* click on any `login with {provider}`
* close the popped up tab/page without granting access
* refresh main page
* you'll see that iframe with `/private_data` resetted request

In logs you can see:
```
auth-example    | 2018/12/29 04:56:12 [REST] GET - /private_data - 192.168.48.1 - 200 (0) - 195.9µs
auth-example    | 2018/12/29 04:56:12 http: panic serving 192.168.48.1:59624: runtime error: invalid memory address or nil pointer dereference
auth-example    | goroutine 42 [running]:
auth-example    | net/http.(*conn).serve.func1(0xc000140dc0)
auth-example    |       /usr/local/go/src/net/http/server.go:1746 +0xd0
auth-example    | panic(0x82f9e0, 0xbf8f90)
auth-example    |       /usr/local/go/src/runtime/panic.go:513 +0x1b9
auth-example    | github.com/go-pkgz/auth/middleware.(*Authenticator).auth.func2.1(0x922fc0, 0xc0001a3dc0, 0xc0001c6d00)
auth-example    |       /go/pkg/mod/github.com/go-pkgz/auth@v0.0.0-20181229073233-4ee872edfbc1/middleware/auth.go:88 +0x6ad
auth-example    | net/http.HandlerFunc.ServeHTTP(0xc000069920, 0x922fc0, 0xc0001a3dc0, 0xc0001c6d00)
auth-example    |       /usr/local/go/src/net/http/server.go:1964 +0x44
auth-example    | github.com/go-chi/chi.(*ChainHandler).ServeHTTP(0xc0000aeac0, 0x922fc0, 0xc0001a3dc0, 0xc0001c6d00)
auth-example    |       /go/pkg/mod/github.com/go-chi/chi@v3.3.3+incompatible/chain.go:31 +0x52
auth-example    | github.com/go-chi/chi.(*Mux).routeHTTP(0xc0000826c0, 0x922fc0, 0xc0001a3dc0, 0xc0001c6d00)
auth-example    |       /go/pkg/mod/github.com/go-chi/chi@v3.3.3+incompatible/mux.go:424 +0x239
[...]

Issue #1, opened 2018-12-28
Add an example service with both minimal backend and frontend


umputun:
The goal is to make a demo service doing something very simple but demonstrating end-to-end usage of the library.

I'm going to make top-level `example` directory for this and will have it as a fully automated build producing docker image. The final container with everything in will run somewhere in DO.

@Reeywhaar - how about UI side, can I count on you? Need smth really trivial, maybe a couple of "pages" to demonstrate access to protected resources with login via all providers. If you willing to help I'll give you more details.

Reeywhaar:
@umputun okay, I'm in, nothing hard here, I guess.

umputun:
thx. yeah, should be really trivial thingy

umputun:
@Reeywhaar all yours.

Under `_example` there is an ... example. `docker-compose build && docker compose up` to start the thing. It will listen on http://127.0.0.1:8080 to "/public" and "/private" (both generated by the server).  Private rejects till you do auth. i.e. http://127.0.0.1:8080/auth/dev/login?from=/private

All API (just a few calls) described in [API section](https://github.com/go-pkgz/auth#api)

I just need an example of SPA demoing smth with protected resource `/private_data`. Something simple but not too ugly, up to you. And speaking of ugly - if you can also de-uglify the [login form](https://github.com/go-pkgz/auth/blob/master/provider/dev_provider.go#L179) I made for dev auth server - it will be very nice of you.

one more thing - this web app should do XSRF header trick, i.e. send back `X-XSRF-TOKEN` header with cookie value from XSRF-TOKEN.

Reeywhaar:
@umputun got it, already there. I'll ping if something not clear.

PR #3, opened 2018-12-29
Example frontend

Reeywhaar:
Made simple page: [screenshot](https://user-images.githubusercontent.com/884649/50537476-21ee8300-0b71-11e9-83f2-ea40f8c24688.png)

I decided to show private_page as just an iframe, so it doesn't duplicate user info.

I have already [said](https://github.com/go-pkgz/auth/issues/1#issuecomment-450482781) that we have trouble with `token/auth` naming issue. Made an issue #4 

Opened issue #2

Also, I decided not to store `XSRF-TOKEN` cookie anywhere, I think it has no point for example, what do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,9 @@ This library provides "social login" with Github, Google, Facebook and Yandex.
 
 - Multiple oauth2 providers can be used at the same time
 - Special `dev` provider allows local testing and development
-- JWT stored in a secure cookie and with XSRF protection. Cookies can be session-only
+- JWT stored in a secure cookie with XSRF protection. Cookies can be session-only
 - Minimal scopes with user name, id and picture (avatar) only
-- Integrated avatar proxy with FS, boltdb or gridfs storage
+- Integrated avatar proxy with FS, boltdb and gridfs storages
 - Support of user-defined storages for avatars
 - Black list with user-defined validator
 - Multiple aud (audience) supported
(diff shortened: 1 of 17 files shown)


Reeywhaar:
About XSRF-TOKEN: I meant that I don't save this cookie value neither in localStorage, nor making it a non-session cookie, so basically, when user close/open browser the XSRF-TOKEN cookie will be missing, but JWT present. Yes right now it has no effect, but I don't know what behaviour would be if xsrf-protection will be on. But, yes, I send it back via headers, as you mentioned:
https://github.com/Reeywhaar/auth/blob/d91d5a77689e6c2cec142be8f51862bf772ae02e/_example/frontend/main.js#L10-L23

About duplication: I meant that I decided to show this page as just an Iframe, and not making it like a table, like I do with `/auth/user` data:
screen shot 2018-12-30 at 01 42 02
Not duplication, but UI similarity and fact that both `/auth/user` and `/private_data` basically contain same data.

Reeywhaar:
How it should be re-obtained when missing? I don't understand what state will be when user make a call to `/auth/user` without XSRF-TOKEN but with JWT? Will it be set with response automatically, or should I make additional request to set this cookie?

umputun:
those cookies should be always paired and both of them set by the server during the login, at the same time. Lack of  XSRF-TOKEN will cause reject after I remove `DisableXSRF:true` flag. Nothing special should be done to re-obtained, just the usual login.

Reeywhaar:
Yeah, my point was to store XSRF-TOKEN in localStorage, to keep sessions between browser reloads. Am I understand right that it expected behaviour to login each time user reloads browser?

umputun:
No, this is not the expected behavior. By default, both cookies are not session cookies and if you reopen the browser they should be in. UI can alter this behavior by sending `&session=1` in login's query params. In this case, both cookies will be session-only and should disappear after browser's restart.

pls note - there is a bug in my refresh logic (just noticed) which may ignore session status and make refresehd JWT cookie non-sessional.

Reeywhaar:
Oh, I see, I was confused and thought XSRF-TOKEN is session only unlike JWT.

Reeywhaar:
Added little more commits (sorry for rebase mess), everything seems working fine.

umputun:
Looks good, thx!

PR #5, opened 2018-12-29
Dev Auth layout

Reeywhaar:
Made dev auth layout. M?
screen shot 2018-12-29 at 16 47 02


umputun:
not exactly like this in safari

fyka3-201812-29124141-qsqnp

PR #6, opened 2018-12-29
Fix dev oauth

Reeywhaar:
To https://github.com/go-pkgz/auth/pull/5#issuecomment-450512078
The layout is completely broken, because of go Sprintf interpolation. Percent marks in css caused the issue.
I don't know should I fix it myself, maybe you know how make it better, but anyway.


umputun:
@Reeywhaar I have changed to template, so no need for the conflicting fix

Issue #8, opened 2018-12-31
Web example doesn't send XSRF header on /private_data request


umputun:
@Reeywhaar  - I have enabled XSRF check and see this:

<img width="1015" alt="kj8ex_20181231_011637 run1p" src="https://user-images.githubusercontent.com/535880/50556052-d4ba0080-0c99-11e9-9a47-117a9f05f0c0.png">

for sure it was rejected: `2018/12/31 01:14:38 [DEBUG] auth failed, can't get token: xsrf mismatch` and "Protected page contents:" shows Unauthorized

umputun:
all good, thx

<img width="878" alt="jv6ja_20181231_034450 d2r7g" src="https://user-images.githubusercontent.com/535880/50557821-78151080-0cae-11e9-8b19-6ac80050908e.png">

PR #9, opened 2018-12-31
fix iframe xsrf header missing

Reeywhaar:
Fix for #8 

Fix a bit hacky, because there is no way to send headers with an iframe. Have to make ajax call, then create object URL from received data and set url of an iframe to said object URL.

### DIFF
--- a/_example/frontend/index.html
+++ b/_example/frontend/index.html
@@ -5,7 +5,7 @@
 	<meta name="viewport" content="width=device-width, initial-scale=1.0">
 	<meta http-equiv="X-UA-Compatible" content="ie=edge">
 	<title>GO-PKGZ/AUTH</title>
-	<link rel="stylesheet" href="./style.css">
+	<link rel="stylesheet" href="./style.css?v=2">
 </head>
 <body>
 	<header class="main-header">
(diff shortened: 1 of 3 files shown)


umputun:
I don't even understand why we need iframe here and then have to resolve issues caused by iframe presence. Maybe just get rid of this completely? If it means a little bit duplication of the code (I think this was the reason for iframe) - so be it.

Anyhow, if you think this how things usually done in ui world - fine with me. I just don't want this example to be tricky or clever.

Reeywhaar:
Nah, it's not tricky, nor clever, it just not, hm, real iframe?
Personally I don't think we need iframe neither, it just was your requirement to show somehow `/private_data`, but basically it's the same data we have in `/auth/user` plus timestamp.
So, remove it?

umputun:
/private_date is not the same as /user. I just didn't have a good idea what unique to return, so just added ts to what I had, but the idea was to demonstrate our ability to hit protected URL and get some data back. Not sure why it even matter what the content of the response if we just show json directly.

umputun:
yeah, just show this json somehow. I had it as <pre> with formatted json.

Reeywhaar:
Check it out.
screen shot 2018-12-31 at 12 04 58

By the way, I noticed that if I login with dev provider under name different from `dev_user`, I'm unable to get contents of `/private_data`. It returns 401 unauthorized.

umputun:
this is exactly as it should be :) https://github.com/go-pkgz/auth/blob/master/_example/backend/main.go#L44

just an example of blocking, it allows dev_ users only and marke dev_admin as admin with some extra attributes

PR #13, opened 2019-01-23
Enable Yandex Open Auth

VaultVulp:
Right now Yandex Open Auth can not be enabled, due to the fact that it is replaced with the `NewFacebook` Open Auth instance. I thought this wasn't intentional because `NewYandex` struct exists in a corresponding source file.

Yet again I wasn't able to provide additional test-cases. But I ran existing tests (with exception of coveralls) successfully.


umputun:
thx

PR #19, opened 2019-04-09
Fix: use custom cookies and headers if provided

minchik:


umputun:
Thx. This was the real bug and the fix is good. However, I don't like what massive test change making them all using custom names. I think,  in 99% of the time default values will be used and I'd like tests to reflect it. In other words, we probably want an extra test (i.e. `TestJWT_NotDefault`) doing set/get for non-default values and keep existing tests with defaults.

minchik:
You are right! These tests look not as good as tests with default names. But there is a good reason to use custom names for all the tests. When we use custom names we are 100% sure that the code is valid for default names too (as long as TestJWT_NewDefault passes) but this statement is not valid vice versa. Tests with default parameters are just code duplication if I add additional tests with none default parameters. They will always fail as long as tests with none default parameters fail. Current tests can also eliminate possible bugs if you add additional checks in the GET method. 
Nevertheless, if you insist I'll change it.

umputun:
makes sense

PR #20, opened 2019-04-10
Simplify claims validation

zonescape:
I have simplified claims validation a bit.

Note also that in the current version of the code this line:
https://github.com/go-pkgz/auth/blob/f3439d6075ac8d5bb6658cc71eb70edd86758336/token/jwt.go#L170
adds `jwt.ValidationErrorExpired` flag if it is not present.

PS
Why expired tokens are allowed?

### DIFF
--- a/token/jwt.go
+++ b/token/jwt.go
@@ -159,20 +159,12 @@ func (j *Service) Parse(tokenString string) (Claims, error) {
 	return *claims, j.validate(claims)
 }
 
+// allow expired token
 func (j *Service) validate(claims *Claims) error {
-	cerr := claims.Valid()
+	c := *claims // shallow copy
+	c.ExpiresAt = 0
 
-	if cerr == nil {
-		return nil
-	}
-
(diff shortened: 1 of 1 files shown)


umputun:
> adds jwt.ValidationErrorExpired flag if it is not present.

I don't see how this a problem. The goal of this validation is to suppress `jwt.ValidationErrorExpired` and return false only in one case. However, for the general sanity, It will be better to return the original error, i.e. `cerr`.

> Why expired tokens are allowed?

For refresh functionality.

I don't like the change for three reasons:
- simplified version looks more confusing, even it simpler. It doesn't communicate intent well enough and for this reason, you needed a comment. It also raises the question "am I setting/changing the original claim here"  (the answer is no, but still some mental effort).  In all areas remotely related to security between verbose and smart I'll prefer verbose if it makes the snippet clear.
- simplified version doesn't solve any problem and doesn't extend functionality. This is refactoring of a "good enough" code.
- you don't seem to use the package (based on the question why such token even allowed) and generally, it makes me question your changes even more.

PR #22, opened 2019-04-11
Claims validation - preserve orginal error

zonescape:
The latest [commit](https://github.com/go-pkgz/auth/commit/2db2238aa8ff296a23d29eb300955bdb866bbd32) effectively does nothing because `e` is a pointer.


umputun:
LGTM

PR #28, opened 2019-07-10
Proposed changes to address: /auth prefix hardcoded in providers

### LINKED ISSUE #7: /auth prefix hardcoded in providers
The example and documentation show mount like this: `router.Mount("/auth", authRoutes) ` however attempt to change "/auth" path to smth else not really supported due to provider's callback urls.

This may cause confusion and should be addressed somehow.

cxww107:
To resolve #7 

login/logout/callback url is always formed (and hardcoded) as:
ROOT_URL/{router path}/PROVIDER/{login | callback | logout}

Would it make sense to define redirect URL in Oauth2Handler Login method in runtime as:
p.conf.RedirectURL = p.URL + r.URL.Path + "/callback" ?

### DIFF
--- a/provider/oauth2.go
+++ b/provider/oauth2.go
@@ -21,13 +21,12 @@ type Oauth2Handler struct {
 	Params
 
 	// all of these fields specific to particular oauth2 provider
-	name        string
-	redirectURL string
-	infoURL     string
-	endpoint    oauth2.Endpoint
-	scopes      []string
-	mapUser     func(userData, []byte) token.User // map info from InfoURL to User
-	conf        oauth2.Config
+	name     string
+	infoURL  string
+	endpoint oauth2.Endpoint
+	scopes   []string
+	mapUser  func(userData, []byte) token.User // map info from InfoURL to User
+	conf     oauth2.Config
 }
 
 // Params to make initialized and ready to use provider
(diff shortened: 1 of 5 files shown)


provider/oauth2.go
```diff
@@ -110,6 +109,11 @@ func (p Oauth2Handler) LoginHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	elems := strings.Split(r.URL.Path, "/")
```
umputun: I'd suggest moving this to a method, i.e. `(p Oauth2Handler) makeRedirURL(path string)`. This will allow to get the url from other place

provider/oauth2.go
```diff
@@ -61,13 +61,12 @@ func initOauth2Handler(p Params, service Oauth2Handler) Oauth2Handler {
...
 		ClientID:     service.Cid,
 		ClientSecret: service.Csecret,
-		RedirectURL:  service.redirectURL,
 		Scopes:       service.scopes,
 		Endpoint:     service.endpoint,
 	}
 
 	p.Logf("[DEBUG] created %s oauth2, id=%s, redir=%s, endpoint=%s",
-		service.name, service.Cid, service.endpoint, service.redirectURL)
+		service.name, service.Cid, service.endpoint)
```
umputun: if redir url composer in another method (see my other comment) will be nice to restore here in debug, as practically setting up provider very often fail due to unexpected redir

umputun:
Good idea, thx. I have added a couple of minor suggestions, but generally it is all good.

provider/oauth2.go
cxww107: As we intend to use handler route to make redir url, at initialization point  it is not known what routing path will be used to serve auth handler.  But we can show the pattern of RedirectURL in log. 
E.g. redir=http://127.0.0.1:8080/YOUR_ROUTE/google/callback

PR #29, opened 2019-07-28
Add support of custom oauth2 server (issue#11)

### LINKED ISSUE #11: Add support of custom oauth2 server
The goal is to allow passing of [go-oauth2/oauth2](https://github.com/go-oauth2/oauth2) as a special kind of provider, start the server and route request accordingly. Probably it will also allow removing the current `dev` provider (will be replaced by go-oauth2/oauth2 based one).

cxww107:
To resolve #11 

I used dev_provider as example for this implementation. Basically I just replaced hardcoded parts of dev_provider with corresponding methods of go-oauth2/oauth2.


provider/oauth2.go
```diff
@@ -161,6 +161,8 @@ func (p Oauth2Handler) AuthHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	p.conf.RedirectURL = p.makeRedirURL(r.URL.Path)
```
cxww107: go-oauth2/oauth2 does check of redir url also during token exchange

provider/custom_server.go
```diff
@@ -0,0 +1,306 @@
...
+	}()
+
+	err := c.httpServer.ListenAndServe()
+	c.Logf("[WARN] dev oauth2 server terminated, %s", err)
+}
+
+func (c *CustomOauthServer) handleAuthorize(w http.ResponseWriter, r *http.Request) {
+	// Called for first time, ask for username
+	if c.WithLoginPage && (r.ParseForm() != nil || r.Form.Get("username") == "") {
+		userLoginTmpl, err := template.New("page").Parse(customLoginTmpl)
```
umputun: Test coverage for this file is below 70% and this branch of code (within `if c.WithLoginPage && ...`) not covered at all.

_Not sure if you can see coveralls report, but in case if it open publicly see https://coveralls.io/builds/24836939/source?filename=provider%2Fcustom_server.go#L80_

cxww107: Yes, I can see coveralls report. I had to check test coverage before submitting pull request. I'll take care of it.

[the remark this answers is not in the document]
cxww107: I added function `LoginPageHandler`  to allow user to handle login page on its own. Usage example in here `_example/custom/main.go` 
Let me know if you prefer more stricter way to let pass just a template.

umputun:
the build failed due to lint issue. They are probably minor, but will be nice to get them addressed:

```
auth.go:318:9: nilness: impossible condition: nil != nil (govet)
980	if err != nil {
981	       ^
982provider/custom_server.go:67:5: shadow: declaration of "err" shadows declaration at line 44 (govet)
983				err := c.OauthServer.HandleAuthorizeRequest(w, r)
984				^
985provider/custom_server.go:73:5: shadow: declaration of "err" shadows declaration at line 44 (govet)
986				err := c.OauthServer.HandleTokenRequest(w, r)
987				^
988provider/custom_server.go:78:9: shadow: declaration of "err" shadows declaration at line 44 (govet)
989				ti, err := c.OauthServer.ValidationBearerToken(r)
990				    ^
991provider/custom_server.go:147:25: `retrive` is a misspelling of `retrieve` (misspell)
992		p.Logf("[ERROR] can't retrive domain from service URL %s", p.URL)
993		                      ^
994auth.go:298:32: `retrive` is a misspelling of `retrieve` (misspell)
995		return fmt.Errorf(`failed to retrive client from custom ouath server 
996		                             ^
997auth.go:303:45: `diffrent` is a misspelling of `different` (misspell)
998		s.logger.Logf("[WARN] client domain=%s is diffrent from service root url=%s", client.GetDomain(), s.opts.URL)
999		                                          ^
1000auth.go:327:32: `retrive` is a misspelling of `retrieve` (misspell)
1001		s.logger.Logf("[ERROR] can't retrive domain from service URL %s", s.opts.URL)
1002		                             ^
1003provider/custom_server.go:41:1: cyclomatic complexity 17 of func `(*CustomOauthServer).Run` is high (> 15) (gocyclo)
1004func (c *CustomOauthServer) Run(ctx context.Context) {
1005^
1006provider/custom_server.go:80:36: ST1013: should use constant http.StatusUnauthorized instead of numeric literal 401 (stylecheck)
1007					http.Error(w, http.StatusText(401), 401)
```

cxww107:
Lint errors you mentioned are fixed with [d22b9ef](https://github.com/go-pkgz/auth/pull/29/commits/d22b9ef4a9624c52ba89c2f6eb8c91bd3ce31c10)  and [bea3585](https://github.com/go-pkgz/auth/pull/29/commits/bea35852011147a354610e119fc8072dcd317412)

provider/custom_server.go
```diff
@@ -0,0 +1,306 @@
...
+
+func (c *CustomOauthServer) handleUserInfo(w http.ResponseWriter, r *http.Request) {
+	ti, err := c.OauthServer.ValidationBearerToken(r)
+	if err != nil {
+		http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
+		return
+	}
+	userID := ti.GetUserID()
+
+	ava := fmt.Sprintf(c.Domain+":%d/avatar?user=%s", custOAuthPort, userID)
```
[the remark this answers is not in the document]
cxww107: [Fixed](https://github.com/go-pkgz/auth/pull/29/commits/950b10c41713a7160266eef533c28097c61785cd)

umputun:
oops, sorry my bad. It is 1am here, and after a day of heavy codding reading inversed order of commits got me confused.

umputun:
A question about `_example/custom`. It looks like you work inside of the GOPATH and, unless I miss something, this example is not "go mod" friendly. Could you add  go.mod and go.sum pls, i.e. "go mod init" + "go get"? Probably you will need to use the `replace` for `github.com/go-pkgz/auth`, in the same way as [_example/go.mod](https://github.com/go-pkgz/auth/blob/master/_example/go.mod#L3) does.

umputun:
Another suggestion regarding the example. Currently, it is a little bit confusing for someone running it as `go run main.go`. After hitting `http://127.0.0.1:8080/auth/custom/login` it gets prompt and with admin/admin authenticates just fine. However attempt to hit protected URL in the browser, i.e. `http://127.0.0.1:8080/private` still returns "Unauthorized" due to understandable xsrf mismatch. The example doesn't print much either (opts.Logger not assigned), so it can be not easy to figure what's going on.

I'd suggest moving the custom provider example to top-level (example/main.go). This way, we will get it nicely integrated and easily demonstratable as it will be shown as one more provider in addition to others.

umputun:
A few things:

1. The example missing dependencies in go.mod/go.sum. In order to add them you can do `go get`. If you work inside of GOPATH it will be `GO111MODULE=on go get`
1. As you used dev_provider,which is special/dev thingy, as an example, the CustomProvider doesn't follow the usual logic all normal providers implement, i.e. it is not added to service explicitly with `service.AddCustomProvider`
1. Generally speaking, custom provider exposes too many implementation details to the consumer. I mean, the only goal to have `auth.go` is to help the user with the wiring of all things together and hide (enclose) dependency initialization. For example, it doesn't expose jwtService but provides high-level wrapper as a part of auth.NewService. The same level of abstraction needed for the custom provider as well. Ideally, the user should fill CustomProviderOpt and pass it to `service.AddCustomProvider("custom123", copts)` and it will do the job.

cxww107:
I got it. Thank you for your remarks. 
sorry for go mod. I forgot again that _example/main.go has its own go.mod

cxww107:
As you suggested I added the method to register custom provider in handler stack: service.AddCustomProvider("custom123", copts)
Where copts are endpoint, infoURL and mapUserFn. The idea is allow to add the custom provider running on a different host. As we wouldn’t  sure about implementation of mapUserFn on server side I allowed to inject one. 

After I removed this part from auth.go it is now responsibility of library user to initiate and start custom server. I added a helper method to initiate server instance and prefill some options needed later for AddCustomProvider. Tbh Im not sure about this. There are obvious things in this method but maybe you’ll find it useful. 

I changed a bit the names as well trying to differentiate the process of registering provider from starting oauth server.

umputun:
A few suggestions and observations. Pls note, I'm not 100% sure those are good suggestions, just smth to think of:

1. `NewCustomServer(srv *goauth2.Server, sopts CustomServerOpts) (*CustomServer, CustomProviderOpt)` - accepting options and returning another option(s) from the constructor, which is highly unusual. Can CustomProviderOpt be encapsulated as a private field in `CustomServer`?
1. In case if it can, `NewCustHandler` could be converted to a method of CustomServer and renamed to `CustHandler` or even `Handler`.
1. Maybe `AddCustomProvider(name string, copts provider.CustomProviderOpt)` wrapper should accept `CustomServer` instead of `CustomProviderOpt`?

cxww107:
@umputun 
1,2 are valid points. I have to refactor the code to look more usual and understandable.  

Regarding 3:
I just try to clarify what I tried to achieve. 
I wanted this method is to be more general. I'd like `AddCustomProvider` be able to register not only go-oauth2/oauth2 provider but maybe some different Oauth2 provider. For example, currently we don't have an implementation of bitbucket login in go-pkgz/auth. But library user would able to add bitbucket login on its own: 
```
service.AddCustomProvider("bitbucket", provider.CustomProviderOpt{
	Endpoint: oauth2.Endpoint{
		AuthURL:  "https://bitbucket.org/site/oauth2/authorize",
		TokenURL: "https://bitbucket.org/site/oauth2/access_token",
	},
	InfoURL: "https://api.bitbucket.org/2.0/user/",
	Cid:     "cid",
	Csecret: "csecret",
	MapUserFn: func(data provider.UserData, _ []byte) token.User {
		userInfo := token.User{
			ID:   "bitbucket_" + token.HashID(sha1.New(), 
                                data.Value("username")),
			Name: data.Value("nickname"),
		}
		return userInfo
	},
	Scopes: []string{"account"},
	})
```
Note: it is not fully testable right now. I have to implement some changes (e.g. make UserData type public). But locally I played a bit with it and it looks good.
Please let me know what do you think about this idea.

umputun:
>  be able to register not only go-oauth2/oauth2 provider but maybe some different Oauth2 provider

makes sense. I have missed that use case, but it will be very nice to have.

cxww107:
1. as you suggested I encapsulated handler opts as a field in CustomServer
2. the method `NewCustHandler` was renamed `NewCustom`. The purpose of this method the same as of methods `NewYandex`, `NewFacebook` etc. 
3. I added two examples in `_example/main.go`. the one for our self-hosted goauth2/oauth2 server, the other for unknown oauth2 provider (i.e. bitbucket) 

I believe Travis checks have failed because right now coveralls is on maintenance.

umputun:
Looks good to me. Thx for the amazing PR, appreciate it!

umputun:
I just realized we completely missed and documentation. I have added smth minimal to README, but feel free to extend/modify it.

cxww107:
@umputun yes, you are right. I'll be back with a new PR for documentation maybe on the weekend. 

Thanks for your advices and remarks. I do not work with Golang on daily basis. it is nice way to learn smth new. Very appreciate it!

umputun:
@nbys due to known security-related issues in go-oauth2/oauth2 I have switched it to v4. The api for v4 is almost identical and the only thing I had to adjust was an extra param to `generates.NewJWTAccessGenerate`.

However, I'm not sure if it works properly. I get all kind of errors as I try to auth with _example. 

Can you pls take a look ?

nbys:
> @nbys due to known security-related issues in go-oauth2/oauth2 I have switched it to v4. The api for v4 is almost identical and the only thing I had to adjust was an extra param to `generates.NewJWTAccessGenerate`.
> 
> However, I'm not sure if it works properly. I get all kind of errors as I try to auth with _example.
> 
> Can you pls take a look ?

I'll take care of this.  To understand, do the errors occur only for `CustomProvider`?

PR #33, opened 2019-08-22
Support twitter auth 

### LINKED ISSUE #31: Support twitter auth
Twitter's auth is a different beast, but it will be nice to have it integrated

cxww107:
To resolve #31

### DIFF
--- a/provider/oauth1.go
+++ b/provider/oauth1.go
@@ -1,0 +1,194 @@
+package provider
+
+import (
+	"context"
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/dghubble/oauth1"
+	"github.com/dgrijalva/jwt-go"
+	"github.com/go-pkgz/auth/logger"
+	"github.com/go-pkgz/auth/token"
+	"github.com/go-pkgz/rest"
+)
+
+// Oauth1Handler implements /login, /callback and /logout handlers for oauth1 flow
+type Oauth1Handler struct {
+	Params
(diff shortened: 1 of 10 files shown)


umputun:
very nice, thx!

From first glance, it looks good and clean, but I'll take a closer looks tonight (hopefully). 

Have you tried to run it with the real twitter auth?

cxww107:
yeap, I tried to run `_example/main.go` with the real twitter application. It looks good. 

There is one thing I am not sure about this implementation though. It is usage of `State` field of handshake. Basically there is no concept of state in oauth1. But there are `request token` and `request secret`. At first they are used for user authentication request. After redirection to callback url (if user authentication was successful)  they are needed for `access token` request. 
To me the purpose of `requestSecret` in oauth1 is somehow similar to `state` in oauth2: to check if after redirection it is still the same user that originated authentication process.

provider/oauth1.go
```diff
@@ -0,0 +1,194 @@
...
+	// use requestSecret as a state in oauth2
+	cid, err := randToken()
+	if err != nil {
+		rest.SendErrorJSON(w, r, h.L, http.StatusInternalServerError, err, "failed to make claim's id")
+		return
+	}
+
+	claims := token.Claims{
+		Handshake: &token.Handshake{
+			State: requestSecret,
```
cxww107: here `requestSecret` is set for later usage in `AuthHandler`

umputun:
> To me the purpose of requestSecret in oauth1 is somehow similar to state in oauth2

makes sense.

Issue #30, opened 2019-08-02
Common shared secret


umputun:
Currently, there is a fundamental limitation in a way secret used. Even with [secret reader](https://github.com/go-pkgz/auth/blob/master/token/jwt.go#L52) injected into the token package, it is impossible to use different secrets for different aud values.

In some complex use cases having different secrets will be desirable. For example, if some service issuing tokens for various services or a different group of users, it may be useful to have different keys (secrets) used for withing each group to sigin and verify token.

Currently, it is impossible to do because aud is a part of the claim, and to extract it from jwt without disabling signature validation, this key should be known **before the parsing step**. In other words, we can [make tokens](https://github.com/go-pkgz/auth/blob/master/token/jwt.go#L98) with a custom (per aud) secret, but we can't [parse](https://github.com/go-pkgz/auth/blob/master/token/jwt.go#L130) them without the real secret.

I don't like the idea of the pre-parsing token without signature verification (`jwt.ParseUnverified` can do it) to extract aud first. I'm not sure what issues it may cause, but it feels unusual and (potentially) not secure.

I am trying to figure an alternative method of achieving the goal, i.e. ability to have a secret per aud.

cxww107:
I am not sure if I got you right. But I present you a draft of implementation. there are some points for discussion #35

PR #35, opened 2019-08-24
add support of aud specific secret

nbys:
to implement #30 

Basically I suggest to introduce a new cookie just for audience. For this cookie will be used common shared secret. After audience is obtained it is possible to use audience-specific secrets. I change implementation of `AudienceReader` to return not only an audience but a secret for it. 

Unfortunately I didn't come up with smth better than break back comparability of `Parse` method. I obviously need to know during the parsing whether common-shared-secret or  audience secret should be used. I see already one flaw with this implementation: audience secret should be provider-specific as well.

Let me know if you think that I am looking in the right direction.

### DIFF
--- a/_example/main.go
+++ b/_example/main.go
@@ -67,6 +67,13 @@ func main() {
 		}),
 		Logger:      log.Default(), // optional logger for auth library
 		UseGravatar: true,          // for verified provider use gravatar service
+		AudienceReader: token.AudienceFunc(func() (map[string]string, error) {
+			return map[string]string{
+				"admins":     "admins-secret",
+				"powerusers": "pu-secret",
+				"users":      "users",
+			}, nil
+		}),
 	}
 
 	// create auth service
--- a/auth.go
+++ b/auth.go
@@ -47,6 +47,7 @@ type Opts struct {
 	DisableIAT  bool // disable IssuedAt claim
 
 	// optional (custom) names for cookies and headers
+	AudCookieName  string // default "AUD-TOKEN"
 	JWTCookieName  string // default "JWT"
 	JWTHeaderKey   string // default "X-JWT"
 	XSRFCookieName string // default "XSRF-TOKEN"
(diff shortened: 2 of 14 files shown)


cxww107:
@umputun as I mentioned there is only a draft for now. I silenced some tests but some failed. I fix if we stick with this approach.

umputun:
I see several problems with the proposed solution:

1. Generally, I don't like to double the number of cookies we produce/consume. In the modern, crazy, post-GDPR world, each cookie is a potential problem.
1. Broken compatibility seems to be the issue for practical use cases. Not the major one, but still
1. It is unclear how this will work with non-cookie jwt exchange. Probably will need to send an extra header, which is odd/unusual
1. I suspect (not sure) it won't work with the cases we use tokens for other things, like verified provider accepting user input. By "won't work," I mean the case if we want to use different secrets per aud.
1. The need of common shared secret is not fully eliminated, just moved to aud token. However, one of the reasons why we want separate secrets is to avoid the situation with compromised shared secret affecting everything. Obviously, in this case, leaked shared secret won't be such a terrible thing as it will alow to produce valid aud tokens only. 

Maybe pre-parse token (without validation) and extract aud is not such a bad idea I thought? Technically, implementation will be much more straightforward, token will remain self-contained, and we won't need any breaking change.

_note: I took just a quick look on the proposal, so maybe some of my conclusions are not fully accurate_

cxww107:
>  Generally, I don't like to double the number of cookies we produce/consume. In the modern, crazy, post-GDPR world, each cookie is a potential problem.
>  The need of common shared secret is not fully eliminated, just moved to aud token. 

you are right. proposed solution is raising more questions than it resolves.

> Maybe pre-parse token (without validation) and extract aud is not such a bad idea I thought? 

I took a look at [PR](https://github.com/dgrijalva/jwt-go/pull/152) for `ParseUnverified` function. It seems that idea behind this method was basically to solve the problems like we have. If we have verification of signature right after `Aud` is extracted, then we are ok? Or do you still see possible attacks?

umputun:
>  If we have verification of signature right after Aud is extracted, then we are ok? Or do you still see possible attacks?

This is the point where I'm not really sure. Theoretically, by allowing pre-parsing without a signature verification we allow anything the potential attacker passes on to be processed. I don't see any obviose attack on our auth part if all we do is getting value for aud without signature check (with `ParseUnverified`). My concern is different - is it possible to cause some issues to jwt library by sending some specially prepared input? 

_The concern I've listed may not be a real one, just over-paranoid me ;)_

cxww107:
>  My concern is different - is it possible to cause some issues to jwt library by sending some specially prepared input?

just a thought. if we can't answer this question with 100% assurance, maybe we can try to minimize possibility of danger by some degree. for example not to use the same `parser` instance for unverified parsing and verification. maybe some checks before we use jwt library? _(but not sure what those checks could be)_

PR #46, opened 2020-04-26
Update project

paskal:
- bump ci configuration
- stricter golangci-lint configuration
- bump go modules


_example/frontend/main.js
```diff
@@ -12,8 +12,7 @@ function req(endpoint, data = {}) {
   const token = cookies["XSRF-TOKEN"];
 
   if (cloneData.hasOwnProperty("headers")) {
-    const headersClone = new Headers(cloneData.headers);
```
umputun, line 15: I have no clue what was the meaning of const and if your removal is justified

paskal: It's not used anywhere, that's why IDE suggested to remove it.

@akellbl4 would you be so kind as to review frontend changes? It's all in single commit, here is [the link](https://github.com/go-pkgz/auth/pull/46/commits/db2f9920d84e084795ef1e93f12b67aff9d0f478).

akellbl4:
@paskal all good in these changes.

PR #48, opened 2020-05-14
Example: according to doc string /private_data returns json with user info

vblz:
Also fix obsolete paths to example


umputun:
thx

Issue #50, opened 2020-05-19
Proporsal: way to modify Claims during auth process


vblz:
What do you think about something like `ClaimsUpd`, but for auth process, for tokens received from clients?

When it could be useful: modifying `Claims` by internal logic.
For example it would be convenient to give admins rights for users with ID from list.

But on the other hand adding this feature will separate token received from client and `Claims` presentation, that's why this might be not really good idea.

umputun:
`CalimsUpd` allows modifying token at the time of generation and refresh. Here [the practical example](https://github.com/umputun/remark42/blob/master/backend/app/cmd/server.go#L868) with ClaimsUpd used to set admin with some internal logic.

The only reason to have similar functionality on the level of `token.ValidatorFunc` (called via auth middleware on each request) I can think of, is to alter claims generated by an external system, not issued but the same producer. I can imagine such cases for inter-service communication, i.e. on service issuing token and hits API of another service, and this another service has to alter the claim somehow. However, all cases I have in mind are better to be handled on the issuer level (if the claim should have admin:true the issuer should make it like this in the first place).

Another problem with this idea is a potential roundtrip to the store on each and every request. Probably can be described in readme, but still, one more place to make it wrong.

vblz:
I solved my issue by wrapping middleware into another middleware, so there is another way to do that.

It had a case described above, there was already finished service and it was easier to make it on our side than another.

And what about a potential roundtrip to the store every request, there is the exactly same situation with `token.ValidatorFunc`, isn't it?

As long there is workaround with wrapping middleware, may be there is no sense in that proposal. So it is at your discretion the issue might be closed.

umputun:
I actually like the idea to provide such functionality as a part of the middleware. Technically speaking, the scope of this claims update has very little to do with auth/tokens, but rather a part of some application-related logic, so middleware seems to be the right place. Another advantage - if we make it as a middleware user could insert updater into particular routes only.

Wanna submit PR with this?

PR #51, opened 2020-05-24
Adding middleware to modify UserInfo in request context

vblz:
Closes go-pkgz/auth#50

### DIFF
--- a/middleware/userUpdater.go
+++ b/middleware/userUpdater.go
@@ -1,0 +1,37 @@
+package middleware
+
+import (
+	"github.com/go-pkgz/auth/token"
+	"net/http"
+)
+
+// UserUpdater defines interface adding extras or modifying UserInfo in request context
+type UserUpdater interface {
+	Update(claims token.User) token.User
+}
+
+// UserUpdFunc type is an adapter to allow the use of ordinary functions as UserUpdater. If f is a function
+// with the appropriate signature, UserUpdFunc(f) is a Handler that calls f.
+type UserUpdFunc func(user token.User) token.User
+
+// Update calls f(user)
+func (f UserUpdFunc) Update(user token.User) token.User {
+	return f(user)
+}
+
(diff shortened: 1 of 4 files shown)


middleware/userUpdater_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+func TestUserUpdate_WithoutAuth(t *testing.T) {
+	a := makeTestAuth(t)
+	mux := http.NewServeMux()
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(201)
+	})
+	upd := UserUpdFunc(func(user token.User) token.User {
+		// this method should not be called because there is no user info in request context
+		assert.True(t, false)
```
umputun, line 53: nit: this is a confusing test. I'd rather do `t.Fatal("should not be called without auth")`

middleware/userUpdater.go
```diff
@@ -0,0 +1,37 @@
...
+func (f UserUpdFunc) Update(user token.User) token.User {
+	return f(user)
+}
+
+// UpdateUser update user info with UserUpdater if it exists in request's context. Otherwise do nothing.
+// should be places after either Auth, Trace or AdminOnly middleware.
+func (a *Authenticator) UpdateUser(upd UserUpdater) func(http.Handler) http.Handler {
+	f := func(h http.Handler) http.Handler {
+		fn := func(w http.ResponseWriter, r *http.Request) {
+			user, err := token.GetUserInfo(r)
```
umputun, line 27: I think this can be simplified, i.e. `if user, err := ...; err == nil {...}`

middleware/userUpdater_test.go
vblz: That looks much better, fixed

middleware/userUpdater.go
vblz: fixed

umputun:
LGTM, thx

PR #54, opened 2020-06-17
Add Telegram login

### LINKED ISSUE #49: Telegram Login
see https://github.com/umputun/remark42/issues/707 for details

nbys:
Fixes #49 

I used rather different approach than you describe in the issue. I tried to implement telegram login to be very close to implementation of other oauth2 providers. Main differences between oauth2 and telegram provider are:
1. In oauth2 providers login handler redirect client to login page. In telegram provider login handler renders template with iframe with telegram button
2. `state` is checked on integrity in oauth2 providers between login and auth handler. 
In telegram provider there is custom check of hash with so-called data-check-string

This is working version that I successfully tested with `_example/main.go`. But I haven't added tests, documentation, comments yet. I'd like to hear your opinion before I make final pull request.

### DIFF
--- a/_example/main.go
+++ b/_example/main.go
@@ -88,6 +88,9 @@ func main() {
 		}),
 	)
 
+	// test telegram login
+	service.AddTelegram(os.Getenv("AEXMPL_TELEGRAM_TOKEN"))
+
 	// run dev/test oauth2 server on :8084
 	go func() {
 		devAuthServer, err := service.DevAuth() // peak dev oauth2 server
--- a/auth.go
+++ b/auth.go
@@ -282,7 +282,8 @@ func (s *Service) AddVerifProvider(name, msgTmpl string, sender provider.Sender)
 	s.authMiddleware.Providers = s.providers
 }
 
-func (s *Service) AddTelegram(token string) {
+// AddTelegram adds telegram as auth provider
+func (s *Service) AddTelegram(botToken string) {
 	p := provider.Params{
 		URL:         s.opts.URL,
 		L:           s.logger,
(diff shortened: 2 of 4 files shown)


umputun:
I'm not sure how iframeTemplate will be integrated into real-life flow. Can you provide screenshots, pls?

nbys:
TODO:
- [ ] add tests to new telegram provider
- [ ] add instruction how to use new provider in README

nbys:
1. New login link: 
Screenshot 2020-06-18 at 09 23 38
2. After click new tab with iframe appears:
Screenshot 2020-06-18 at 09 23 53

3. After click on widget:
Screenshot 2020-06-18 at 09 24 16

4. 
Screenshot 2020-06-18 at 09 25 18

umputun:
the flow is really confusing with this extra-step opening iframe, which makes it very different from any other real oauth2 providers. I'm not even sure how this can be integrated into real-life systems.

nbys:
I would prefer to open popup with telegram authorization (screenshot 3) directly. I tried this as well but unfortunately without widget I couldn't get callback working (due to telegram unique flow for auth). Agree, one step more, one click more. it may be confusing.
I don't quite understand your concerns regarding integration into real-life application but maybe I just don't see the whole picture. 
But it was worth to try :)

nbys:
@umputun you mentioned this pull request on podcast. I would like to clarify one point. Actually after user has pressed Accept button (screenshot 3), iframe magic redirects to callback url and other browser tabs will be closed. No need for user to close opened popups. 

I understand your point that it makes no sence to include this workflow to general-purpose oauth library. Nevertheless I decided to use this workaround in my project. I am happy enough with fork, but obviously I would prefer use this repo instead. Would you approve one more public method to `Service` to append self-implemented handlers. smth like that:
```go
func (s *Service) AppendProvider(name string, provider provider.Provider) {...}
```

umputun:
Having AppendProvider in addition to the current AddProvider, Add[Custom|Direct|Verif]Provider is going to be confusing. I have no problem with the idea, but we need a better name for this

Issue #55, opened 2020-06-17
Add Microsoft provider


umputun:
This is going to be another social provider for Microsoft / Azure

PR #56, opened 2020-06-18
Microsoft (azure) oauth2 provider

umputun:
Implements azure aouth2. Handles basic user info as well as a separate authorized call required by azure to retrieve avatar. 

Note: the call hits beta API because production one is not supported yet, see https://github.com/microsoftgraph/microsoft-graph-docs/issues/3990

#55


umputun:
fix #55

PR #57, opened 2020-06-22
Add a POST request handler to the direct provider

o1egl:
In some case it's required to send user provided input data in POST request. This PR adds `application/x-www-form-urlencoded` and `application/json` encoded body processors to the direct handler.

### DIFF
--- a/provider/direct.go
+++ b/provider/direct.go
@@ -2,15 +2,22 @@ package provider
 
 import (
 	"crypto/sha1"
-	"errors"
+	"encoding/json"
+	"mime"
 	"net/http"
 	"time"
 
 	"github.com/dgrijalva/jwt-go"
 	"github.com/go-pkgz/rest"
+	"github.com/pkg/errors"
 
 	"github.com/go-pkgz/auth/logger"
 	"github.com/go-pkgz/auth/token"
+)
+
+const (
+	// MaxHTTPBodySize defines max http body size
+	MaxHTTPBodySize = 1024 * 1024
 )
 
 // DirectHandler implements non-oauth2 provider authorizing user in traditional way with storage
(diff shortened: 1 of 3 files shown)


provider/direct.go
```diff
@@ -13,6 +16,13 @@ import (
 	"github.com/go-pkgz/auth/token"
 )
 
+const (
+	// MaxHTTPBodySize defines max http body size
+	MaxHTTPBodySize = megabyte
```
umputun, line 21: `MaxHTTPBodySize = 1024*1024` will be easier to read

provider/direct.go
```diff
@@ -37,21 +47,48 @@ func (f CredCheckerFunc) Check(user, password string) (ok bool, err error) {
 	return f(user, password)
 }
 
+// Credentials holds user credentials
+type Credentials struct {
```
umputun, line 51: I don't think this struct supposed to be exposed. If so, making it `credentials` makes more sense

provider/direct.go
```diff
@@ -37,21 +47,48 @@ func (f CredCheckerFunc) Check(user, password string) (ok bool, err error) {
...
+// application/json body example:
+// {
+//   "user": "name",
+//   "passwd": "xyz",
+//   "aud": "bar",
+// }
 func (p DirectHandler) LoginHandler(w http.ResponseWriter, r *http.Request) {
-	user, password := r.URL.Query().Get("user"), r.URL.Query().Get("passwd")
-	aud := r.URL.Query().Get("aud")
+	if r.Body != nil {
```
umputun, line 77: probably, moving this body wrapping to getCredentials will make this part closer to the consumer and more readable

PR #65, opened 2020-08-09
Simple Role Based access control

kleash:
- `auth.RBAC` method for handlers
- developers can set optional roles in claims updater `claims.User.SetRole`

Not using attributes to store roles to maintain backward compatibility so that it won't break for anyone whole already have a `role` key in attributes.

### DIFF
--- a/middleware/auth.go
+++ b/middleware/auth.go
@@ -7,6 +7,7 @@ package middleware
 import (
 	"crypto/subtle"
 	"net/http"
+	"strings"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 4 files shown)


middleware/auth.go
```diff
@@ -187,3 +187,22 @@ func (a *Authenticator) basicAdminUser(r *http.Request) bool {
...
+// this handler internally wrapped with auth(true) to avoid situation if RBAC defined without prior Auth
+func (a *Authenticator) RBAC(role string, next http.Handler) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		user, err := token.GetUserInfo(r)
+		if err != nil {
+			http.Error(w, "Unauthorized", http.StatusUnauthorized)
+			return
+		}
+
+		if role != user.Role {
```
umputun, line 201: maybe EqualFold will be a better one here? Not 100% sure, but case sensetivty in roles probably not needed

kleash:
It's working fine though, only test is failing in my local.

> Looking good, thx. A minor suggestion about role matching.
> 
> Another thing missing - some description/mention for readme

Added lit bit description in README

PR #66, opened 2020-08-17
RBAC with multiple roles and chains-friendly

umputun:
Converted RBAC middleware to a more traditional signature and added support of matching against multiple roles.

The reason for this change - in any modern framework supporting middleware chains (i.e. route.With(middleware, middleware2, ....) the old signature expecting handler as a part of the call is not smth usual and not easy to use. In addition, in practical use cases, one route (or group of routes) often allowed for multiple roles.

Another thing included in this PR - old test with ClaimUpdater counted on token's refresh and didn't work in normal circumstances. I'm not really sure how it even passed Actions CI, probably due to incorrect (different) clock on action workers. Replaced by pre-generated token and simplified test's code.

@kleash pls take a look. This change won't be backward compatible but I think it will be the right thing to do.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ func main() {
 - `middleware.Auth` - requires authenticated user
 - `middleware.Admin` - requires authenticated admin user
 - `middleware.Trace` - doesn't require authenticated user, but adds user info to request
-- `middleware.RBAC` - requires authenticated user with passed role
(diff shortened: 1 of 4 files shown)


kleash:
Looks good, I was going to open PR for multiple roles, but this covers everything well.

Issue #73, opened 2021-01-06
New opts to pass same site attribute


janikvonrotz:
The SameSite attribute is missing.

About the attribute: https://golang.org/pkg/net/http/#SameSite

Originally I submitted this issue: https://github.com/umputun/remark42/pull/841
It would set the attribute in https://github.com/go-pkgz/auth/blob/2f012acddab465e12498267f563089ccf458492a/token/jwt.go#L240

However, setting it statically is bad practice and the attribute should be introduced as an option to https://github.com/go-pkgz/auth/blob/2f012acddab465e12498267f563089ccf458492a/token/jwt.go#L52 

Thinking about something like `opts.SameSite`.

Can anybody help implementing this?

PR #74, opened 2021-01-06
support same site attribute #73

umputun:
add same site option #73


umputun:
@janikvonrotz - I have implemented #73, pls take a look just in case.

janikvonrotz:
@umputun Awesome! Thanks for the implementation. Nothing to add here.

PR #82, opened 2021-04-09
Add custom basic auth checker

zebox:
Would you consider proposal to add feature which implement user custom basic auth checker.

I tried wrote tests which covered all my changes, but can't add integration test because it required add additional input parameter for `prepService`. By default prepService defined `AdminPasswd` if I add hard `BasicAuthChecker` to `perpService` options it will brake `TestIntegrationBasicAuth` test. After test server started change service options unavailable (after call `prepService` return). If I can been have opportunity send service options to `prepService` as input parameter i can done integration test, but it require change func input signature for all call in test file.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -66,11 +66,12 @@ type Opts struct {
 	AvatarRoutePath   string       // avatar routing prefix, i.e. "/api/v1/avatar", default `/avatar`
 	UseGravatar       bool         // for email based auth (verified provider) use gravatar service
 
-	AdminPasswd    string                  // if presented, allows basic auth with user admin and given password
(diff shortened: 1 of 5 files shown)


umputun:
the code looks clean and likely will do the right thing. However I don't get the point of the BasicAuth interface. First of all it has very little to do with the token package, second I'd rather declare it near the consumer, but all of this irrlevant as I don't see why this should be even an interface/struct and not just a passed function

zebox:
I tried saved similarity of implementation with direct auth provider, because one almost like with basic auth. But I agree with you about excess using interface/struct.  I can put the `basicAuthCheker` type implementation in to `auth.go` file and define one as func type. What are you think about that way of implementation?

PR #84, opened 2021-04-16
add basicAuth description to readme

zebox:


README.md
```diff
@@ -383,13 +383,34 @@ It will run fake aouth2 "server" on port :8084 and user could login with any use
...
+   //...
+}
+```
+
 ### Other ways to authenticate
 
 In addition to the primary method (i.e. JWT cookie with XSRF header) there are two more ways to authenticate:
 
 1. Send JWT header as `X-JWT`. This shouldn't be used for web application, however can be helpful for service-to-service authentication.
 2. Send JWT token as query parameter, i.e. `/something?token=<jwt>`
```
umputun, line 413: I'd suggest keeping `3. Basic access authentication` and put just something like `see [Basic Basic authentication](link) above`

zebox: Ok. I will change it. Does `Basic authentication` section stay here or move it below the `Other way.. `?

Issue #61, opened 2020-06-28
Add support of "Sign In with Apple"


umputun:
This seems to be oauth2 with some extras.  See https://developer.apple.com/sign-in-with-apple/

This [article](https://developer.okta.com/blog/2019/06/04/what-the-heck-is-sign-in-with-apple) adds some explanation as well.

fr3fou:
any news?

PR #86, opened 2021-05-10
added sign in with apple

zebox:
Implementation of `Sign in with Apple`  #61

Apple API has some limit for implement full features compatible with the package:

1. Apple hasn't API for fetch avatar. If any one know how to do it let me know (also no `userInfo` api).

2. Fetch user name (if set in scope) possible only one, at first user login with a service. With next login  `userName` field doesn't exist in API response, until user will delete sign with an app in account profile (security section).

I will planning update the `Readme` after code review.

Waiting for your comments

### DIFF
--- a/README.md
+++ b/README.md
@@ -465,9 +465,11 @@ Then add an Apple provider that accepts the following parameters:
 ```
 
 **Limitation:**
-* Map a userName (if specific scope defined) can be only at first login for a user.
-Every next login with Apple, under account which sign in early, will no return field with userName and provider can't fetch the name until a user delete sign in for you service with Apple ID in  Apple account profile (security section).
-Provider always get user `UID` (`sub` claim) and `email` (if email scope defined) from claims of `IDToken`.
+* Map a userName (if specific scope defined) is only sent in the upon initial user sign up.
+Subsequent logins to your app using Sign In with Apple with the same account do not share any user info and will only return a user identifier in IDToken claims.
+This behaves correctly until a user delete sign in for you service with Apple ID in own Apple account profile (security section). 
+It is recommend that you securely cache the at first login containing the user info for bind it with a user UID at next login.
+Provider always get user `UID` (`sub` claim) in `IDToken`.
 
 * Apple doesn't have an API for fetch avatar and user info.
 
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # auth - authentication via oauth2, direct and email
 [![Build Status](https://github.com/go-pkgz/auth/workflows/build/badge.svg)](https://github.com/go-pkgz/auth/actions) [![Coverage Status](https://coveralls.io/repos/github/go-pkgz/auth/badge.svg?branch=master)](https://coveralls.io/github/go-pkgz/auth?branch=master) [![godoc](https://godoc.org/github.com/go-pkgz/auth?status.svg)](https://pkg.go.dev/github.com/go-pkgz/auth?tab=doc)
 
-This library provides "social login" with Github, Google, Facebook, Microsoft, Twitter, Yandex, Battle.net and Telegram as well as custom auth providers and email verification.
+This library provides "social login" with Github, Google, Facebook, Microsoft, Twitter, Yandex, Battle.net, Apple and Telegram as well as custom auth providers and email verification.
 
 - Multiple oauth2 providers can be used at the same time
 - Special `dev` provider allows local testing and development
@@ -421,7 +421,59 @@ By default, this library doesn't print anything to stdout/stderr, however user c
 
 Authentication handled by external providers. You should setup oauth2 for all (or some) of them to allow users to authenticate. It is not mandatory to have all of them, but at least one should be correctly configured.
 
-#### Google Auth Provider
+#### Apple Auth Provider
+To configure this provider, a user requires an Apple developer account (without it setting up a sign in with Apple is impossible). [Sign in with Apple](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api) lets users log in to your app using their two-factor authentication Apple ID.
+
+Follow to next steps for configuring on the Apple side:
+
+1. Log in [to the developer account](https://developer.apple.com/account).
+1. If you don't have an App ID yet, [create one](https://developer.apple.com/account/resources/identifiers/add/bundleId). Later on, you'll need **TeamID**, which is an "App ID Prefix" value.
+1. Enable the "Sign in with Apple" capability for your App ID in [the Certificates, Identifiers & Profiles](https://developer.apple.com/account/resources/identifiers/list) section.
+1. Create [Service ID](https://developer.apple.com/account/resources/identifiers/list/serviceId) and bind with App ID from the previous step. Apple will display the description field value to end-users on sign-in. You'll need that service **Identifier as a ClientID** later on**.**
+1. Configure "Sign in with Apple" for created Service ID. Add domain where you will use that auth on to "Domains and subdomains" and its main page URL (like `https://example.com/` to "Return URLs".
+1. Register a [New Key](https://developer.apple.com/account/resources/authkeys/list) (private key) for the "Sign in with Apple" feature and download it. Write down the **Key ID**. This key will be used to create [JWT](https://developer.apple.com/documentation/sign_in_with_apple/generate_and_validate_tokens#3262048) Client Secret.
+1. Add your domain name and sender email in the Certificates, Identifiers & Profiles >> [More](https://developer.apple.com/account/resources/services/configure) section as a new Email Source.
+
+After completing the previous steps, you can proceed with configuring the Apple auth provider. Here are the parameters for AppleConfig:
+
+- _ClientID_ (**required**) - Service ID identifier which is used for Sign with Apple
+- _TeamID_ (**required**) - Identifier a developer account (use as prefix for all App ID)
+- _KeyID_ (**required**) - Identifier a generated key for Sign with Apple
+
+
+```go
+    // apple config parameters
+	appleCfg := provider.AppleConfig{
+		TeamID:   os.Getenv("AEXMPL_APPLE_TID"), // developer account identifier
+		ClientID: os.Getenv("AEXMPL_APPLE_CID"), // service identifier
+		KeyID:    os.Getenv("AEXMPL_APPLE_KEYID"), // private key identifier
+	}
+```
+
+Then add an Apple provider that accepts the following parameters:
+* `appleConfig (provider.AppleConfig)`  created above
+* `privateKeyLoader (PrivateKeyLoaderInterface)`
+
+`PrivateKeyLoaderInterface` [implements](https://github.com/go-pkgz/auth/blob/master/provider/apple.go#L98:L100) a loader for the private key (which you downloaded above) to create a `client_secret`. The user can use a pre-defined function `provider.LoadApplePrivateKeyFromFile(filePath string)` to load the private key from local file.
+
+`AddAppleProvide` tries to load private key at call and return an error if load failed. Always check error when calling this provider.
+
+```go
+    if err := service.AddAppleProvider(appleCfg, provider.LoadApplePrivateKeyFromFile("PATH_TO_PRIVATE_KEY_FILE")); err != nil {
+		log.Fatalf("[ERROR] failed create to AppleProvider: %v", err)
+	}
+```
+
+**Limitation:**
+* Map a userName (if specific scope defined) can be only at first login for a user.
+Every next login with Apple, under account which sign in early, will no return field with userName and provider can't fetch the name until a user delete sign in for you service with Apple ID in  Apple account profile (security section).
+Provider always get user `UID` (`sub` claim) and `email` (if email scope defined) from claims of `IDToken`.
+
+* Apple doesn't have an API for fetch avatar and user info.
+
+See [example](https://github.com/go-pkgz/auth/blob/master/_example/main.go#L83:L93) before use.
+
+#### Google Auth Provide
 
 1.  Create a new project: https://console.developers.google.com/project
 2.  Choose the new project from the top right project dropdown (only if another project is selected)
--- a/auth.go
+++ b/auth.go
@@ -257,6 +257,26 @@ func (s *Service) AddDevProvider(port int) {
 	s.providers = append(s.providers, provider.NewService(provider.NewDev(p)))
 }
 
+// AddAppleProvider allow SignIn with Apple ID
+func (s *Service) AddAppleProvider(appleConfig provider.AppleConfig, privKeyLoader provider.PrivateKeyLoaderInterface) error {
+	p := provider.Params{
+		URL:         s.opts.URL,
+		JwtService:  s.jwtService,
+		Issuer:      s.issuer,
+		AvatarSaver: s.avatarProxy,
+		L:           s.logger,
+	}
+
+	// Error checking at create need for catch one when apple private key init
+	appleProvider, err := provider.NewApple(p, appleConfig, privKeyLoader)
+	if err != nil {
+		return errors.Wrap(err, "an AppleProvider creating failed")
+	}
+
+	s.providers = append(s.providers, provider.NewService(appleProvider))
+	return nil
+}
+
 // AddCustomProvider adds custom provider (e.g. https://gopkg.in/oauth2.v3)
 func (s *Service) AddCustomProvider(name string, client Client, copts provider.CustomHandlerOpt) {
 	p := provider.Params{
--- a/provider/apple.go
+++ b/provider/apple.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"crypto"
 	"crypto/ecdsa"
+	"crypto/sha1"
 	"crypto/x509"
 	"encoding/json"
 	"golang.org/x/oauth2"
@@ -176,7 +177,7 @@ func NewApple(p Params, appleCfg AppleConfig, privateKeyLoader PrivateKeyLoaderI
 		mapUser: func(claims jwt.MapClaims) token.User {
 			var usr token.User
 			if uid, ok := claims["sub"]; ok {
-				usr.ID = fmt.Sprintf("apple_%s", uid.(string))
+				usr.ID = "apple_" + token.HashID(sha1.New(), uid.(string))
 			}
 			return usr
 		},
@@ -317,7 +318,7 @@ func (ah AppleHandler) AuthHandler(w http.ResponseWriter, r *http.Request) {
 		rest.SendErrorJSON(w, r, ah.L, http.StatusInternalServerError, err, "exchange failed")
 		return
 	}
-	ah.Logf("[DEBUG]response data %+v", resp)
+	ah.Logf("[DEBUG] response data %+v", resp)
 	if resp.Error != "" {
 		rest.SendErrorJSON(w, r, ah.L, http.StatusInternalServerError, nil, fmt.Sprintf("fetch IDtoken response error: %s", resp.Error))
 		return
@@ -326,7 +327,7 @@ func (ah AppleHandler) AuthHandler(w http.ResponseWriter, r *http.Request) {
 	// trying to fetch Apple public key (JWK) for verify token signature, it need for verify IDToken received from Apple
 	keySet, err := fetchAppleJWK(r.Context(), ah.conf.jwkURL)
 	if err != nil {
-		ah.L.Logf("failed to fetch JWK from Apple key service: " + err.Error())
+		ah.L.Logf("[ERROR] failed to fetch JWK from Apple key service: " + err.Error())
 		rest.SendErrorJSON(w, r, ah.L, http.StatusInternalServerError, nil, fmt.Sprintf("failed to fetch JWK from Apple key service: %s", resp.Error))
 		return
 	}
@@ -335,7 +336,7 @@ func (ah AppleHandler) AuthHandler(w http.ResponseWriter, r *http.Request) {
 	tokenClaims := jwt.MapClaims{}
 	_, err = jwt.ParseWithClaims(resp.IDToken, tokenClaims, keySet.keyFunc)
 	if err != nil {
-		ah.L.Logf("failed to get claims: " + err.Error())
+		ah.L.Logf("[ERROR] failed to get claims: " + err.Error())
 		rest.SendErrorJSON(w, r, ah.L, http.StatusInternalServerError, nil, fmt.Sprintf("failed to token validation, key is invalid: %s", resp.Error))
 		return
 	}
@@ -348,10 +349,8 @@ func (ah AppleHandler) AuthHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// try parse user name data if it exist in scope and service response
-	if jUser != "" {
-		ah.parseUserData(&u, jUser)
-	}
+	// try parse username if one exist at response or noname assign
+	ah.parseUserData(&u, jUser)
 
 	cid, err := randToken()
 	if err != nil {
@@ -489,9 +488,11 @@ func (ah *AppleHandler) parseUserData(user *token.User, jUser string) {
 
 	// Catch error for log only. No need break flow if user name doesn't exist
 	if err := json.Unmarshal([]byte(jUser), &userData); err != nil {
-		ah.L.Logf("[ERROR] failed to parse user data %s: %v", user, err)
+		ah.L.Logf("[DEBUG] failed to parse user data %s: %v", user, err)
+		user.Name = "noname_" + user.ID[6:12] // paste noname if user name failed to parse
 		return
 	}
+
 	user.Name = fmt.Sprintf("%s %s", userData.Name.FirstName, userData.Name.LastName)
 }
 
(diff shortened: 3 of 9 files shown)


umputun:
thx for the pr

I have not reviewed the code yet but two things noticed right away – linter complaints (some are valid) and lack of any documentation.

zebox:
Ofcourse, you are right. PR is wet yet. I will fix linter compilation errors (golintci why was it disabled in my IDE🤷‍♂️) and add information how to use the apple  provider in to readme.

README.md
```diff
@@ -420,8 +420,66 @@ By default, this library doesn't print anything to stdout/stderr, however user c
 ## Register oauth2 providers
 
 Authentication handled by external providers. You should setup oauth2 for all (or some) of them to allow users to authenticate. It is not mandatory to have all of them, but at least one should be correctly configured.
+#### Apple Auth Provider
+For configure this provider user require Apple developer account (without it setting up sign in with Apple impossible). [Sign in with Apple](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api) lets users log in to your app using their two-factor authentication Apple ID.
 
-#### Google Auth Provider
+Follow to next steps for configure on Apple side: 
+
+1. At first you need login in to developer account and enable capabilities `Sign in with Apple" for specific App ID (in Certificates, Identifiers & Profiles section ).  Also you need `App ID Prefix` as TeamID value, it's required for configure Apple auth provider.
```
paskal, line 428: Is it possible to add links for instruction steps?

README.md
```diff
@@ -420,8 +420,66 @@ By default, this library doesn't print anything to stdout/stderr, however user c
 ## Register oauth2 providers
 
 Authentication handled by external providers. You should setup oauth2 for all (or some) of them to allow users to authenticate. It is not mandatory to have all of them, but at least one should be correctly configured.
+#### Apple Auth Provider
```
paskal, line 423:
```suggestion

#### Apple Auth Provider
```

README.md
```diff
@@ -420,8 +420,66 @@ By default, this library doesn't print anything to stdout/stderr, however user c
...
+    if err := service.AddAppleProvider("apple", appleCfg, provider.LoadApplePrivateKeyFromFile(os.Getenv("AEXMPL_APPLE_PRIVKEY_PATH"))); err != nil {
+		log.Fatalf("[ERROR] create AppleProvider failed: %v", err)
+	}
+```
+**Limitation:** 
+* Map a userName (if specific scope defined) can be only at first login for a user. 
+Every next login with Apple, under account which sign in early, will no return field with userName and provider can't fetch user name until a user delete sign in for you service with Apple ID in  Apple account profile (security section).
+Provider always get user `UID` (`sub` claim) and `email` (if email scope defined) from claims of `IDToken`.
+
+* Apple hasn't API for fetch avatar and user info.
```
paskal, line 478:
```suggestion
* Apple doesn't have an API for fetch avatar and user info. 
```

README.md
```diff
@@ -420,8 +420,66 @@ By default, this library doesn't print anything to stdout/stderr, however user c
...
+	}
+```
+**Limitation:** 
+* Map a userName (if specific scope defined) can be only at first login for a user. 
+Every next login with Apple, under account which sign in early, will no return field with userName and provider can't fetch user name until a user delete sign in for you service with Apple ID in  Apple account profile (security section).
+Provider always get user `UID` (`sub` claim) and `email` (if email scope defined) from claims of `IDToken`.
+
+* Apple hasn't API for fetch avatar and user info. 
+
+See [example](https://github.com/zebox/auth/blob/master/_example/main.go#L83:L93) before use.
```
paskal, line 480:
```suggestion
See [example](https://github.com/go-pkgz/auth/blob/master/_example/main.go#L83:L93) before use.
```

provider/apple.go
```diff
@@ -0,0 +1,532 @@
...
+func (ah *AppleHandler) initPrivateKey() error {
+	if ah.PrivateKeyLoader == nil {
+		return errors.New("private key loader interface is nil")
+	}
+	if ah.conf.KeyID == "" {
+		return errors.New("keyID can't be empty")
+	}
+	sKey, err := ah.PrivateKeyLoader.LoadPrivateKey()
+	if err != nil {
+		return err
```
paskal, line 209: Better wrap it with some context, like "problem loading private key", as well as others in that function.

auth.go
```diff
@@ -257,6 +258,26 @@ func (s *Service) AddDevProvider(port int) {
...
+		JwtService: s.jwtService,
+		Issuer:     s.issuer,
+		//AvatarSaver: s.avatarProxy, // not implemented avatar for AppleProvider yet
+		L: s.logger,
+	}
+
+	// Error checking at create need for catch one when apple private key init
+	appleProvider, err := provider.NewApple(name, p, appleConfig, oauth2.Endpoint{}, privKeyLoader)
+	if err != nil {
+		return err
```
paskal, line 274: I propose to wrap it into something like "problem creating an AppleProvider"

provider/apple.go
```diff
@@ -0,0 +1,532 @@
...
+// exchange sends the validation token request and gets access token and user claims
+// (e.g. https://developer.apple.com/documentation/sign_in_with_apple/generate_and_validate_tokens)
+func (ah *AppleHandler) exchange(ctx context.Context, code, redirectURI string, result interface{}) (err error) {
+
+	// check client_secret is expired and recreate new (client_secret JWT) if need
+	ok, err := ah.isClientSecretExpired()
+	if err != nil || !ok {
+		ah.conf.clientSecret, err = ah.createClientSecret()
+		if err != nil {
+			return err
```
paskal, line 389: I propose to wrap error returns in that function as well.

paskal (changes requested):
Just a couple of clarifications. I'll try to test the sign in when I'll have a chance, later on. I have a developer Apple account for that purpose.

README.md
zebox: links added

provider/apple.go
zebox: added error wrap in current in commit [8c4df02 ](https://github.com/go-pkgz/auth/pull/86/commits/8c4df026d051c2a4d6a2260156917e2eb418b92a)

auth.go
zebox: added error wrap in [8f262ce](https://github.com/go-pkgz/auth/pull/86/commits/8f262cef53e491146a7af03b723e7df6972c7eb7) commit

zebox (commented):
Your comments added

provider/apple.go
```diff
@@ -0,0 +1,532 @@
...
+	if err != nil {
+		return false, err
+	}
+	claims, ok := tkn.Claims.(jwt.MapClaims)
+	if !ok {
+		return false, errors.New("can't convert token claims to standard claims")
+	}
+	return claims.VerifyExpiresAt(time.Now().Unix(), true), nil
+}
+func (ah *AppleHandler) parseUserData(user *token.User, jUser string) {
```
paskal, line 486: Please make this function covered with tests, called from the exposed function if possible.

paskal (commented):
Code and doc look good to me, I'll try to test it in reality as soon as possible.

provider/apple.go
```diff
@@ -0,0 +1,534 @@
...
+
+	if name == "" {
+		return nil, errors.New("empty name for apple provider not allowed")
+	}
+
+	if p.L == nil {
+		p.L = logger.NoOp
+	}
+
+	// configuring default values
```
umputun, line 147: are those params should be even configurable? Maybe we just use const here as I don't really see the reason to redefine user agent, auth url and token url

provider/apple.go
```diff
@@ -0,0 +1,534 @@
...
+
+		endpoint: endpoints,
+		mapUser: func(claims jwt.MapClaims) token.User {
+			var usr token.User
+			if uid, ok := claims["sub"]; ok {
+				usr.ID = fmt.Sprintf("apple_%s", uid.(string))
+			}
+
+			// this claim exist if "email" defined in scope
+			if email, ok := claims["email"]; ok {
```
umputun, line 179: this one is for what exactly? Why we even need it and how email_verified (for verified account with the manual confirmation) even related to this?

provider/apple.go
```diff
@@ -0,0 +1,534 @@
...
+	req.Header.Add("accept", AcceptHeader)
+	req.Header.Add("user-agent", ah.conf.UserAgent) // apple requires a user agent
+
+	res, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+
+	// Apple REST API response either 200 (ok OK) or 400 (if error)
+	if res.StatusCode >= 400 {
```
umputun, line 416: maybe we better be more explicit here? i.e. if res.StatusCode == http.StatusBadRequest or != http.StatusOk

provider/apple.go
```diff
@@ -0,0 +1,534 @@
...
+		if err = res.Body.Close(); err != nil {
+			ah.L.Logf("[ERROR] close request body failed when get access token: %v", err)
+		}
+	}()
+	return err
+}
+
+// getTokenClaims decodes the id_token in response for extract claims with user data
+func (ah AppleHandler) getTokenClaims(idToken string) (jwt.MapClaims, error) {
+	tkn, _, err := new(jwt.Parser).ParseUnverified(idToken, jwt.MapClaims{})
```
umputun, line 431: maybe just pass in (inject) token.jwt and use it instead of manipulating directly? Also, it should have very good reason to parse unverified and I'm not sure what the reason for this

provider/apple.go
```diff
@@ -0,0 +1,534 @@
...
+
+// createClientSecret use for create the JWT client secret required to make requests to the Apple validation server.
+// for more details go to link: https://developer.apple.com/documentation/sign_in_with_apple/generate_and_validate_tokens#3262048
+func (ah *AppleHandler) createClientSecret() (string, error) {
+	if ah.conf.privateKey == nil {
+		return "", errors.New("private key can't be empty")
+	}
+	// Create the Claims
+	now := time.Now()
+	exp := now.Add(time.Second * 86400).Unix() // default value
```
umputun, line 452: is it 24 hours? do we really need it that long?

provider/apple.go
```diff
@@ -0,0 +1,534 @@
...
+	}
+	ah.JwtService.Reset(w)
+}
+
+// exchange sends the validation token request and gets access token and user claims
+// (e.g. https://developer.apple.com/documentation/sign_in_with_apple/generate_and_validate_tokens)
+func (ah *AppleHandler) exchange(ctx context.Context, code, redirectURI string, result interface{}) (err error) {
+
+	// check client_secret is expired and recreate new (client_secret JWT) if need
+	ok, err := ah.isClientSecretExpired()
```
umputun, line 385: the whole handling with unverified makes me nervous as I can't see all possible attacks exploiting this. Is it possible to replace by the real verification?

umputun (changes requested):
Thx a lot, impresive work!

I haven't check the actual logic, but some implementation details raised questions.

provider/apple.go
zebox: I'm agree with you about exclude user-agent from config params, but auth and token urls not include to `AppleConfig` and exist only in constructor `NewApple`. It's need for change they for tests only.

umputun: I would rather change it directly in tests if this is the only reason for the exposure if we are talking about same-package unit tests only

provider/apple.go
zebox: it's claim return in token with other claims. I add it as additional flag for user check. If it isn't need i remove it.

provider/apple.go
zebox: Yes. I will fix it.

provider/apple.go
umputun: well, in fact, it exposes email in the final token. And I don't get the part about email_verified, is it smth apple provides?

provider/apple.go
zebox: I will try parse token claims with `private key` instead unverified parse. If it possible I change implementation for this.

provider/apple.go
zebox: No. What are you offer?  1 hour?

provider/apple.go
zebox: You are right! I will change it to verified parse.

provider/apple.go
umputun: maybe even shorter? for the verified provider we do 30m

provider/apple.go
zebox: Yes. Apple return token with this additional claim. Email claim exist as well.
[Token description](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/authenticating_users_with_sign_in_with_apple#3383773)

provider/apple.go
zebox: ok

provider/apple.go
zebox: Ok, will change it

zebox (commented):
all changes done

paskal:
Sincere apologies, still haven't got time to test it with the real account: will try to go trough the doc tomorrow evening.

README.md
```diff
@@ -421,7 +421,65 @@ By default, this library doesn't print anything to stdout/stderr, however user c
...
+
+* _ClientID_ (**required**) - Service ID identifier which use for `Sign with Apple`
+* _TeamID_ (**required**) - Identifier a developer account (use as prefix for all App ID)
+* _KeyID_ (**required**) - Identifier a generated key for `Sign with Apple`
+* _Scopes_ (optional) - The [scope](https://developer.apple.com/documentation/sign_in_with_apple/clientconfigi/3230955-scope) is the amount of user information requested from Apple. Right now, there are only two options name and email. You can request the user’s name or email. You can also choose to request both or neither.
+* _ClientSecretExpire_ (optional) - time in seconds when client secret will expire (default: 1800 secs = 30 minutes)
+
+```go 
+    // apple config parameters  
+	appleCfg := provider.AppleConfig{
```
paskal, line 449: If some keys are required, your code must explicitly fail when they are not provided.

paskal: Nvm, I see that it's doing the check already.

auth.go
```diff
@@ -257,6 +257,26 @@ func (s *Service) AddDevProvider(port int) {
 	s.providers = append(s.providers, provider.NewService(provider.NewDev(p)))
 }
 
+// AddAppleProvider allow SignIn with Apple ID
+func (s *Service) AddAppleProvider(name string, appleConfig provider.AppleConfig, privKeyLoader provider.PrivateKeyLoaderInterface) error {
```
paskal, line 261: Another provider creator is generic and requires a name because of that, I don't think this one needs it. @umputun ?

paskal (commented):
I've rewritten a readme, I hope you don't mind it.

I've followed the instruction and after I create a provider I see an error`[ERROR] create AppleProvider failed: an AppleProvider creating failed: keyID can't be empty` despite having AppleConfig.KeyID defined, I'll try to figure out what's wrong but so far it hasn't passed the smoke test.

paskal (approved):
It works.

provider/apple.go
```diff
@@ -0,0 +1,529 @@
...
+	Error string `json:"error"`
+}
+
+// AppleConfig is the main oauth2 required parameters for "Sign in with Apple"
+type AppleConfig struct {
+	ClientID           string   // the identifier Services ID for your app created in Apple developer account.
+	TeamID             string   // developer Team ID (10 characters), required for create JWT. It available, after signed in at developer account, by link: https://developer.apple.com/account/#/membership
+	KeyID              string   // private key ID  assigned to private key obtain in Apple developer account
+	ClientSecretExpire int64    // time in seconds when client secret will expired (default: 30 minutes)
+	Scopes             []string // for apple provider allow only "email" and "name" scope values
```
umputun, line 77: I would suggest droping this one and enforce the lowest scope possible (name?), like we do with other providers. We also could drop user email  upstream

provider/apple.go
```diff
@@ -0,0 +1,529 @@
...
+	ah := AppleHandler{
+		Params: p,
+		name:   name,
+
+		conf: AppleConfig{
+			ClientID:           appleCfg.ClientID,
+			TeamID:             appleCfg.TeamID,
+			KeyID:              appleCfg.KeyID,
+			ClientSecretExpire: appleCfg.ClientSecretExpire,
+			Scopes:             appleCfg.Scopes,
```
umputun, line 158: see above, unless i missed smth setting here []stting{"name"} will do

provider/apple.go
```diff
@@ -0,0 +1,529 @@
...
+		},
+		mapUser: func(claims jwt.MapClaims) token.User {
+			var usr token.User
+			if uid, ok := claims["sub"]; ok {
+				usr.ID = fmt.Sprintf("apple_%s", uid.(string))
+			}
+
+			// this claim exist if "email" defined in scope
+			if email, ok := claims["email"]; ok {
+				usr.Email = email.(string)
```
umputun, line 174: this may go away if we drop email scope

provider/apple.go
```diff
@@ -0,0 +1,529 @@
...
+
+// createClientSecret use for create the JWT client secret required to make requests to the Apple validation server.
+// for more details go to link: https://developer.apple.com/documentation/sign_in_with_apple/generate_and_validate_tokens#3262048
+func (ah *AppleHandler) createClientSecret() (string, error) {
+	if ah.conf.privateKey == nil {
+		return "", errors.New("private key can't be empty")
+	}
+	// Create the Claims
+	now := time.Now()
+	exp := now.Add(time.Second * 1800).Unix() // default value
```
umputun, line 460: maybe more human-friendy form will be more readbale? lile time.Minute*30

provider/apple.go
```diff
@@ -0,0 +1,529 @@
...
+
+	return authURL.String(), nil
+
+}
+
+func (ah AppleHandler) makeRedirURL(path string) string {
+	elems := strings.Split(path, "/")
+	newPath := strings.Join(elems[:len(elems)-1], "/")
+
+	return strings.TrimRight(ah.URL, "/") + strings.TrimRight(newPath, "/") + urlCallbackSuffix
```
umputun, line 528: TrimRight is not the same as TrimSuffix you probably meant

provider/apple.go
```diff
@@ -0,0 +1,529 @@
...
+			LastName  string `json:"lastName"`
+		} `json:"name"`
+		Email string `json:"email"`
+	}
+
+	var userData UserData
+
+	// Catch error for log only. No need break flow if user name doesn't exist
+	err := json.Unmarshal([]byte(jUser), &userData)
+	if err != nil {
```
umputun, line 495: nit: this likely can be combined to the single statement, i.e. if err := blah ...

provider/apple_pubkeys.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	apk.Modulus = big.NewInt(0).SetBytes(bufferN)
+	apk.Exponent = int(big.NewInt(0).SetBytes(bufferE).Int64())
+
+	return nil
+}
+
+// AppleKeySet sets of Apple public key
+type AppleKeySet struct {
```
umputun, line 152: is this one public for real? Looks like it has private methods only

provider/apple_pubkeys.go
```diff
@@ -0,0 +1,182 @@
...
+	"time"
+)
+
+// appleKeysURL is the endpoint URL for fetch Apple’s public key
+const appleKeysURL = "https://appleid.apple.com/auth/keys"
+
+// ApplePublicKey is the Apple public key object
+// Apple public key is a data structure that represents a cryptographic key as JSON Web Key (JWK)
+// based on RFC-7517 https://datatracker.ietf.org/doc/html/rfc7517
+type ApplePublicKey struct {
```
umputun, line 28: should be public?

provider/apple_pubkeys.go
```diff
@@ -0,0 +1,182 @@
...
+	Usage     string   `json:"use"`
+	Algorithm string   `json:"alg"`
+	Modulus   *big.Int `json:"n"`
+	Exponent  int      `json:"e"`
+
+	publicKey *rsa.PublicKey
+}
+
+// AppleRawKey is raw json object
+type AppleRawKey struct {
```
umputun, line 40: should be public?

umputun (changes requested):
more minor things

auth.go
umputun: yeat, bothered me as well. But i don't think it can be done with the same params as others

provider/apple.go
umputun: I know you got it from oauth2, and it has the same issue. I'll fix oauth2 later on

auth.go
paskal: I think it would be better to hardcode name=“apple” for consistency and not allow user changing it.

umputun: agree

zebox: What is about if a user want add more then one AppleProvider or that case not acceptable? I look to the `TelegramProvider` where a user can define a provider name.
If you think use an AppleProvider without name define by a user,  I will remove name define from AppleProvider constructor.

provider/apple.go
zebox: Ok.  I will change it.

provider/apple.go
zebox: I define in seconds, because assume that a user want to define lifitime in seconds. Are you think remove lifitime redefine from the `AppleProvider` config too?

auth.go
umputun: > more then one AppleProvider

this would be ... unusual. Personally, I have never seen an appincation providing multiple google logins or multiple Facebook logins at the same time. I don't really see in what situation such functionality would be desirable. 

Regarding TelegramProvider - yeah, we should address it for the same reason and hardcode provider's name. I guess I missed this one during review because TG doesn't have a full integration like yours and wired in via a custom provider.

zebox (commented):
ok.

provider/apple.go
umputun: this is just a duration, right? so, 30 minutes seems easier to read. 

> Are you think remove lifitime redefine from the AppleProvider config too?

I think this is a good idea.

auth.go
zebox: Ok.  I will make hardcoded name for AppleProvider.

provider/apple.go
zebox: Ок.  I will remove it.

umputun:
LGTM. could you pls resolve the conflict and rebase those commits?

umputun:
@zebox if you have any question about how to squash it feel free to ping @paskal he recently helped another contributor with the similar change

zebox:
@paskal Could you help me with squash commits? I squashed all commits in my forked repo and push one [commit] (https://github.com/go-pkgz/auth/commit/0aab890b3a333cf2a07524c7606226c8241f42d5), with all changes, to PR. But github shows 34 commits for merge. 
What do I wrong?

umputun:
I think because you are trying to change the history it should be forced push replacing all the other commits. but will see what @paskal suggests

paskal:
Please just abort the merge (git merge --abort) and do git push --force.

zebox:
thx, i will do it tomorrow.

zebox:
@umputun, I noticed for missed block with noname in mapUser function. Do I need to add it to `mapUser` func? 
e.g.
```go
if userInfo.Name == "" {
    userInfo.Name="noname_" + userInfo.ID[0:12]

```

umputun:
> @umputun, I noticed for missed block with noname in mapUser function. Do I need to add it to `mapUser` func?
> e.g.
> 
> ```go
> if userInfo.Name == "" {
>     userInfo.Name="noname_" + userInfo.ID[0:12]
> ```

Yes, pls. And this is especially important due to "no name on the second call" apple's limitation

zebox:
> > @umputun, I noticed for missed block with noname in mapUser function. Do I need to add it to `mapUser` func?
> > e.g.
> > ```go
> > if userInfo.Name == "" {
> >     userInfo.Name="noname_" + userInfo.ID[0:12]
> > ```
> 
> Yes, pls. And this is especially important due to "no name on the second call" apple's limitation

Ok. I will add it.  In Apple provider it should be check in parseUserData func, because i parse user name after `mapUser` called.

zebox:
All changes tested and ready to merge.

umputun (approved):

provider/apple.go
```diff
@@ -0,0 +1,526 @@
...
+
+	authURL, err := url.Parse(ah.endpoint.AuthURL)
+	if err != nil {
+		return "", err
+	}
+
+	query := authURL.Query()
+	query.Set("state", state)
+	query.Set("response_type", "code")
+	query.Set("response_mode", "form_post")
```
akellbl4, line 511: @zebox hi, I was wondering if there was any particular reason to hardcode `response_mode`.

zebox: @akellbl4 hi, according to [documentation](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/incorporating_sign_in_with_apple_into_other_platforms#3332115) `response_mode` has three parameters: `query`, `fragment `, `form_post `. But `query` and `fragment` includes response parameters to the `redirectURI`. I prefer to avoid show explicit response parameters, with sensitive user data, in a browser URL. Therefore, I send it as form data.

It's like for `login` scheme with `GET` method, where credentials send at query URL, but I prefer send in body with `POST`. 
I know that credentials should be hashed at frontend side, before send to backend, but i think better hides this data from explicit view in a browser. I understand that doesn't added some strong security, but  it's hides credentials from access logs (which logs parameters of URL query on a backend side).

Issue #92, opened 2021-08-15
Telegram auth doesn't set "aud" in JWT token after login


Ksinia:
Telegram login method doesn't set "aud" in generated JWT token which prevents it from being used in remark42. Example JWT token with Telegram auth:

```json
{
  "exp": 1629041975,
  "jti": "5dfe954c0d62d41f5722b2cfdce6242a8fe32b4b",
  "iat": 1629041675,
  "iss": "telegram",
  "user": {
    "name": "Ksenia",
    "id": "telegram_d1660f94f72ab7e578766460c7004e95607fb586",
    "picture": "http://127.0.0.1:8080/api/v1/avatar/f6c6df0ddfcae25b1358c87e1af9904208837983.image",
    "attrs": {
      "admin": false,
      "blocked": false
    }
  }
}
```

With anonymous auth:

```json
{
  "aud": "remark",
  "exp": 1629041567,
  "jti": "eafedda197100f969b902ff79a441d72dabaa852",
  "iat": 1629041267,
  "iss": "remark42",
  "user": {
    "name": "test",
    "id": "anonymous_a94a8fe5ccb19ba61c4c0873d391e987982fbbd3",
    "picture": "http://127.0.0.1:8080/api/v1/avatar/c5f426697ad14b7fe76e2b154e89c5b6cbc24b61.image",
    "attrs": {
      "admin": false,
      "blocked": false
    }
  }
}
```

Is it something related with this package @holykol?

PR #93, opened 2021-08-16
Add Audience, ExpiresAt and NotBefore to telegram JWT token

paskal:
Likely solves #92. Btw we don't have any tests for cookies in telegram auth.


umputun (approved):
I'm not sure if this will solve the issue, but PR looks good to me

PR #95, opened 2021-09-01
Patreon OAuth2 provider.

VMAnalytic:
Added Patreon OAuth2 provider.


umputun:
thx for PR. I have not reviewed it yet in details, however, it seems to implement only the oauth2 part and not the thing I mentioned about the "paying status" claim:

> Regarding subscriber status, i.e. detection of paying subscribers - this info is hopefully available via some API and we can publish it to some optional JWT claim, and the consumer can detect it and alter the behavior

I'm not sure if you are planning to address it as a part of the PR or even noticed it.

VMAnalytic:
@umputun 
I've added optional attribute to the token, but few questions arise.
Patreon documentation is not very clear at some moments. I was guided by [this](https://www.patreondevelopers.com/t/checking-if-a-user-is-currently-subscribed-to-a-campaign/152/1) comment. I created separate account subscribed on myself and it works.

But here are my apprehensions:

- will this work correctly if my subscription expires?
- do we need to differentiate different types of subscriptions?

provider/providers.go
```diff
@@ -191,3 +188,54 @@ func NewMicrosoft(p Params) Oauth2Handler {
...
+			userInfo := token.User{}
+
+			uinfoJSON := uinfo{}
+			if err := json.Unmarshal(bdata, &uinfoJSON); err == nil {
+				userInfo.ID = "patreon_" + token.HashID(sha1.New(), userInfo.ID)
+				userInfo.Name = uinfoJSON.Data.Attributes.FullName
+				userInfo.Picture = uinfoJSON.Data.Attributes.ImageURL
+
+				// check if the user is your subscriber
+				if len(uinfoJSON.Data.Relationships.Pledges.Data) > 0 {
```
umputun, line 233: I'm not sure how this supposed to work. It looks like the check is not checking for "my subscriber" but rather making sure the user subscribed to something, which may or may not be the one we care about.

umputun (changes requested):

provider/providers.go
VMAnalytic: This decision was made based on this information at Patreon API documentation.

> This optional parameter will default to users pledges-to-me my-campaign, which fetches user profile information, pledges to your creator, and your creator info. It will be displayed to the user in human-friendly terms when signing in with Patreon. If your client requires the ability to ask for pledges or campaign data of other users (not just your own campaign), please email platform@patreon.com, and we'll do our best to get back to you shortly.

So that users should be "your" subscriber.

umputun: unless I'm missing something, I don't even see it passes my campaign id in any form. How is this even supposed to know which campaign is mine?

VMAnalytic: I may only assume that they have some kind of mapping between the provided access token and client_id.
I can add another subscription to my account and double check this assumption.

umputun: I guess the source of my confusion is the lack of any mention of campaign id in this instruction. I'm not that familiar with Patreon, but as far as I recall under the same account user may have multiple campaigns/subscriptions?

#### Patreon Auth Provider ####
1.	Create a new Patreon client https://www.patreon.com/portal/registration/register-clients
1.  Fill **"App Name"**, **"Description"**, **"App Category"** and **"Author"** for your site
1.  Under **"Redirect URIs"** enter the correct url constructed as domain + `/auth/patreon/callback`. ie `https://example.mysite.com/auth/patreon/callback`
1.  Take note of the **Client ID** and **Client Secret**

VMAnalytic: That confused me too. But as I've mentioned above campaign_id is not needed. I guess that they do some mapping between provided access_token and client_id that was provided during the first OAuth step. And based on this relation they return only pledges that related to your account.

umputun: > I can add another subscription to my account and double check this assumption.

this would be very nice to make sure it works the way we expect

VMAnalytic: Ok. So I I've tested it and received only 1 pledge for my account.

`            "relationships": {
                "campaign": {
                    "data": {
                        "id": "7381928",
                        "type": "campaign"
                    },
                    "links": {
                        "related": "https://www.patreon.com/api/campaigns/7381928"
                    }
                `
If you want I can share client ID / secret for my test account with you.

umputun (approved):
LGTM, thx!

umputun:
could you, pls squash all the comments into one?

VMAnalytic:
> could you, pls squash all the comments into one?

Done.

PR #96, opened 2021-09-02
Direct with custom user id func

umputun:
The direct provider uses just the name (hashed) to make a user ID. In many cases this is exactly what is needed, i.e. login with user/passwd may expect unique user names, however in some tricky cases we may want to change the default behavior. For example, the implementation of an anonymous provider, where no password is even passed and uniqueness on the name can not be expected.

The PR adds support of `UserIDFunc` overriding a part of the id. The prefix (provider name) still mandated but the rest can be whatever user like. It accepts both name and http.Request params, so it is possible to generate id based on the user name and mix in some ip parts.

I'm not sure if we should enforce hashing of `UserIDFunc` result or left it to the caller

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
 
       - name: install golangci-lint and goveralls
         run: |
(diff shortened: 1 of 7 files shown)


umputun:
@paskal - this is related to what we discussed recently about anonymous provider implementation in remark42. Proposed PR trying to address it with minimal efforts, without making a new provider. 

pls take a look as you have a moment

paskal (approved):

PR #99, opened 2021-10-29
Multiple improvements to telegram auth

paskal:
- sync code style with remark42
- improve parseError returned error description
- handle empty response case in BotInfo
- remove unused fields from botInfo object
- remove the unnecessary returned error from Run
- separate cleanup and processUpdates tickers
- increase processUpdates ticker from 1s to 5s
- set expired requests cleanup ticker to 5m
- increase test coverage

### DIFF
--- a/provider/telegram.go
+++ b/provider/telegram.go
@@ -56,7 +56,8 @@ type TelegramAPI interface {
 }
 
 // changed in tests
-var tgPollInterval = time.Second
+var apiPollInterval = time.Second * 5        // interval to check updates from Telegram API and answer to users
+var expiredCleanupInterval = time.Minute * 5 // interval to check and clean up expired notification requests
 
 // Run starts processing login requests sent in Telegram
 // Blocks caller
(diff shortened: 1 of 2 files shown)


provider/telegram.go
```diff
@@ -412,14 +430,12 @@ func (tg *tgAPI) request(ctx context.Context, method string, data interface{}) e
...
-func (tg *tgAPI) parseError(r io.Reader) error {
-	var tgErr = struct {
+func (tg *tgAPI) parseError(r io.Reader, statusCode int) error {
+	tgErr := struct {
 		Description string `json:"description"`
 	}{}
-
-	if err := json.NewDecoder(r).Decode(&tgErr); err != nil {
-		return errors.Wrap(err, "can't decode error")
+	if err := json.NewDecoder(r).Decode(&tgErr); err == nil {
```
umputun, line 437: i liked flipped check more, i.e. err != nil as it seems to be closer the overall code style

provider/telegram_test.go
```diff
@@ -84,7 +84,7 @@ func TestTelegramUnconfirmedRequest(t *testing.T) {
 	tg.LoginHandler(w, r)
 
 	assert.Equal(t, http.StatusNotFound, w.Code, "response code should be 404")
-	assert.Equal(t, `{"error":"request expired"}`+"\n", w.Body.String())
+	assert.Equal(t, `{"error":"request verification failed"}`+"\n", w.Body.String())
```
umputun, line 87: is it expected? the comment is about expired request but it seems like we lost the specific "expired" context

umputun (commented):
I don't see anything worng here, just a few style-related, minor things

provider/telegram.go
```diff
@@ -377,16 +394,17 @@ func (tg *tgAPI) BotInfo(ctx context.Context) (*botInfo, error) {
 
 	err := tg.request(ctx, "getMe", &resp)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to fetch bot info")
+		return nil, err
```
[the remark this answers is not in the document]
paskal: I played around with it in tests and saw that it doesn't make sense to wrap errors here as we wrap them into `return errors.Wrap(err, "failed to fetch bot info")` on the caller level: it results in `failed to fetch bot info: failed to fetch bot info: err` as an end result. That removed wrap adds nothing useful to Telegram API error message.

PR #100, opened 2021-11-06
[telegram] add the ability to process updates from the third source

paskal:
Previously, loop listening for new Telegram bot messages through API updates had to be run within auth, after that change it would be possible to run ProcessUpdate instead to process updates from the loop running elsewhere.

Also, do not send an error response to the person sending the request in case it doesn't start with /start, e.g. for the regular message: new behavior is to just ignore them.

Required for https://github.com/umputun/remark42/pull/1160

### DIFF
--- a/provider/telegram_test.go
+++ b/provider/telegram_test.go
@@ -202,6 +202,71 @@ func TestTelegramLogout(t *testing.T) {
 	assert.Equal(t, time.Time{}, c.Expires)
 }
 
+func TestTelegramHandler_Name(t *testing.T) {
+	tg := &TelegramHandler{ProviderName: "test telegram"}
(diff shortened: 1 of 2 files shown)


provider/telegram_test.go
```diff
@@ -210,63 +275,63 @@ func TestTelegram_TokenVerification(t *testing.T) {
 		BotInfoFunc: botInfoFunc,
 	}
 
-	th, cleanup := setupHandler(t, m)
+	tg, cleanup := setupHandler(t, m)
```
paskal, line 278: This is done for consistency, 3 other places are `tg` and not `th`.

umputun (approved):
LGTM

PR #104, opened 2022-01-30
[telegram] return telegram bot username when Run is not used

paskal:
In Remark42, telegram auth used with its own API updates listener and TelegramHandler.Run is never called, which currently leaves TelegramHandler.username unset and returned empty when Telegram login attempt is made.

This PR changes that, so that username will be always set or error will be returned otherwise.


umputun (approved):
LGTM

PR #108, opened 2022-02-10
fix _example with CustomProvider on go-auth2/oauth2 server

nbys:
The check below fails in _example because in auth options URL defined as `http://localhost:8080`.  It should match. 
https://github.com/go-oauth2/oauth2/blob/c121ece09e9d992ed7c2e5eda4b2cd3120b9ad6e/manage/util.go#L16

#29


umputun (approved):
thx

PR #110, opened 2022-02-19
style(auth): remove spaces

sashamelentyev:


umputun:
I don't get, sorry:

- why this event a thing ? I mean, having an empty line on the top is not a "style violation" as gofmt and goimprort are fine with this
- why such a selective edit? I have a similar style in many other places but you want to change just a few

the bottom line - this makes no sense to me, closing this PR

PR #119, opened 2022-06-03
sanitize both user and address in verified provider

umputun:
this PR sanitizes both user and address prior to confirmation callback.

This was done to prevent untrusted user input to set those fields (mostly user) to something we may send, for example, as an email. I'm not sure if an email's HTML can be attacked with injected JS, but either way, it prevents malicious attempts to make fake emails with rich content

@paskal - this is the provider you are familiar with and we usually use it for email logins. Pls take a look

### DIFF
--- a/provider/verify.go
+++ b/provider/verify.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/go-pkgz/rest"
 	"github.com/golang-jwt/jwt"
+	"github.com/microcosm-cc/bluemonday"
 
 	"github.com/go-pkgz/auth/avatar"
 	"github.com/go-pkgz/auth/logger"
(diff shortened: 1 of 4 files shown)


umputun:
@paskal one more thing to consider - maybe it makes sense to replace `text/template` with `html/template` here? This should help in cases like this, I think.

provider/verify.go
```diff
@@ -205,3 +211,18 @@ Confirmation for {{.User}} {{.Address}}, site {{.Site}}
 
 Token: {{.Token}}
 `
+
+func (e VerifyHandler) sanitize(inp string) string {
+	p := bluemonday.UGCPolicy()
+	res := p.Sanitize(inp)
+	res = template.HTMLEscapeString(res)
+	res = strings.Replace(res, "&amp;", "&", -1)
```
paskal, line 219: Should be ReplaceAll in modern Go.

paskal (approved):

paskal:
> @paskal one more thing to consider - maybe it makes sense to replace `text/template` with `html/template` here? This should help in cases like this, I think.

I think we should replace all three: verify.go, custom_server.go and dev_provider.go.

PR #122, opened 2022-06-22
Support OpenID tokens for getting using information

alek-sys:
Some providers (e.g. Cognito) provide only minimal subset of user attributes in `/userinfo` endpoint, but much more in the ID tokens. That can be helpful to pass user metadata from the provider to the app. This PR adds support of OpenID tokens as a source of user info, as an extension to the OAuth2 flow. 

OpenID flow is only supported for custom providers at the moment and require JWKS URL configuration. Switching existing providers (e.g. `google`) to OpenID is TBD, but probably not required. Using OpenID Connect (`.well-known/openid-configuration` URLs) is out of scope for this PR. 

Apple flow basically is OpenID, with some customisations on how client secret is passed across. So potentially these two flows can be merged together later on.

### DIFF
--- a/provider/dev_provider.go
+++ b/provider/dev_provider.go
@@ -99,42 +99,53 @@ func (d *DevAuthServer) Run(ctx context.Context) { //nolint (gocyclo)
 					email = d.GetEmailFn(d.username)
 				}
 
-				idClaims := map[string]interface{}{
-					// required OpenID claims
-					"iss": "dev-auth",
-					"sub": "%s",
-					"aud": "client-id",
-					"iat": time.Now().Unix(),
-					"exp": time.Now().Add(1 * time.Hour).Unix(),
-
-					// optional OpenID claims
-					"picture":    fmt.Sprintf("http://127.0.0.1:%d/avatar?user=%s", d.Provider.Port, d.username),
-					"given_name": d.username,
-					"email":      email,
-				}
-
-				if d.CustomizeIdTokenFn != nil {
-					idClaims = d.CustomizeIdTokenFn(idClaims)
-				}
-
-				tk := jwt.NewWithClaims(jwt.SigningMethodRS256, jwt.MapClaims(idClaims))
-				tk.Header["kid"] = "dev-auth-key-1"
-				signedTk, err := tk.SignedString(privateKey)
-				if err != nil {
-					d.Logf("[ERROR] failed to sign ID token")
-					w.WriteHeader(http.StatusInternalServerError)
-					return
-				}
-
-				res := fmt.Sprintf(`{
+				res := `{
 					"access_token":"MTQ0NjJkZmQ5OTM2NDE1ZTZjNGZmZjI3",
-					"id_token": "%s",
 					"token_type":"bearer",
 					"expires_in":3600,
 					"refresh_token":"IwOGYzYTlmM2YxOTQ5MGE3YmNmMDFkNTVk",
 					"scope":"create",
 					"state":"12345678"
-					}`, signedTk)
+				}`
+
+				if d.Provider.UseOpenID {
+					idClaims := map[string]interface{}{
+						// required OpenID claims
+						"iss": "dev-auth",
+						"sub": "%s",
+						"aud": "client-id",
+						"iat": time.Now().Unix(),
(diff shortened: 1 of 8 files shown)


provider/oauth2.go
```diff
@@ -4,30 +4,39 @@ import (
...
+	"github.com/pkg/errors"
 	"io"
 	"net/http"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/go-pkgz/rest"
 	"github.com/golang-jwt/jwt"
+	jwtv4 "github.com/golang-jwt/jwt/v4"
```
umputun, line 17: is there a good reason to have both v4 and pre-gomod 3.x.x ? I think v4 was mostly compat with v3, at least i recall migrating some projects to v4 without any significant problem.

provider/oauth2.go
```diff
@@ -4,30 +4,39 @@ import (
...
 
 	// all of these fields specific to particular oauth2 provider
 	name     string
 	infoURL  string
+	jwksURL  string
 	endpoint oauth2.Endpoint
 	scopes   []string
 	mapUser  func(UserData, []byte) token.User // map info from InfoURL to User
 	conf     oauth2.Config
+	keyfunc  jwt.Keyfunc
```
umputun, line 38: maybe make the relation between these to more explicit with internal struct{fn,mutex}?

provider/oauth2.go
```diff
@@ -235,9 +247,108 @@ func (p Oauth2Handler) LogoutHandler(w http.ResponseWriter, r *http.Request) {
...
+		if err != nil {
+			return nil, nil, errors.Wrap(err, "can't load JWKS keys")
+		}
+	}
+
+	claims := jwt.MapClaims{}
+	parser := jwt.Parser{
+		// claims validation is not considering clock skew and randomly failing with iat validation
+		// nbf and exp are validated below
+		SkipClaimsValidation: true,
```
umputun, line 267: I get the idea and the need, but still makes me worry. Maybe there is a way to do it safer somehow without skip validation?

provider/oauth2.go
```diff
@@ -235,9 +247,108 @@ func (p Oauth2Handler) LogoutHandler(w http.ResponseWriter, r *http.Request) {
...
+}
+
 func (p Oauth2Handler) makeRedirURL(path string) string {
 	elems := strings.Split(path, "/")
 	newPath := strings.Join(elems[:len(elems)-1], "/")
 
 	return strings.TrimSuffix(p.URL, "/") + strings.TrimSuffix(newPath, "/") + urlCallbackSuffix
 }
+
+func (p *Oauth2Handler) tryInitJWKSKeyfunc() error {
```
umputun, line 324: I'm not sure what exactly this thing does, but it fills like the goal of this "try" is to set/init p.keyfunc once? If so, how about sync.Once? Will it make the intent more clear?

provider/dev_provider.go
```diff
@@ -74,7 +90,8 @@
 				}
 
 				state := r.URL.Query().Get("state")
-				callbackURL := fmt.Sprintf("%s?code=g0ZGZmNjVmOWI&state=%s", d.Provider.conf.RedirectURL, state)
+				redirectURI := r.URL.Query().Get("redirect_uri")
+				callbackURL := fmt.Sprintf("%s?code=g0ZGZmNjVmOWI&state=%s", redirectURI, state)
```
umputun, line 94: isn't this change will break back compat for dev provider?

umputun (changes requested):
a few, mostly minor concerns. Some could be due to my quick review withour diving deep into the problem area.

provider/oauth2.go
alek-sys: That's what `keyfunc` is using, I don't think they have v3 compatible version anymore. If you are considering upgrading `auth` to `jwt/v4`, I'm happy to do that (as a separate PR) as well.

provider/oauth2.go
alek-sys: Upgrading to `jwt/v4` should solve this issue as well, they added clock skew recently.

tendant:
This feature is very useful. Is there anything needed for merging this PR?

PR #142, opened 2023-01-14
Fix nil ptr panic in auth.go

Zhomart:
It happens when `err=nil` and `claims.User=nil` in Handlers() func.

https://github.com/go-pkgz/auth/blob/master/auth.go#L184

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -181,11 +181,9 @@ func (s *Service) Handlers() (authHandler, avatarHandler http.Handler) {
 			claims, _, err := s.jwtService.Get(r)
 			if err != nil || claims.User == nil {
 				w.WriteHeader(http.StatusUnauthorized)
-				var msg string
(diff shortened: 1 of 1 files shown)


auth.go
```diff
@@ -181,7 +181,13 @@ func (s *Service) Handlers() (authHandler, avatarHandler http.Handler) {
 			claims, _, err := s.jwtService.Get(r)
 			if err != nil || claims.User == nil {
 				w.WriteHeader(http.StatusUnauthorized)
-				rest.RenderJSON(w, rest.JSON{"error": err.Error()})
+				var msg string
+				if err != nil {
+					msg = err.Error()
+				} else {
+					msg = "User is nil"
+				}
```
umputun, line 189: thx. probably we can get rid of else here with the default  msg := "user is nil" and keeping just the if part of err != nil

PR #156, opened 2023-03-04
fix: Get request from web to private page

eddiefisher:
if you have Get request from web to private page, you can't set data in header without JS


umputun:
thx for the PR, however, I don't understand the problem and the proposed solution. To me, it seems like you will effectively disable all XSRF flow by doing this, but if this was the goal, you could already do it by setting `DisableXSRF`. Unless I'm missing something, the change will compare the CSRF token (set in a cookie) with itself but the goal here is to compare a token set by the server with `XSRFCookieName` with the token set by the client via `XSRFHeaderKey`.

eddiefisher:
hmmm... you right

PR #159, opened 2023-04-01
Bump golangci-lint

paskal:
Currently, the pipelines are timing out and update might help fix it.


umputun:
1.20 should be "1.20" or yml will see it as 1.2

PR #166, opened 2023-05-20
Fixed wrong gravatar url generation (lower case, file type)

JayDi85:
Fixed two bugs:
1. According to [gravatar docs](https://en.gravatar.com/site/implement/images/) must be used trimmed and lower case email for url generation;
2. File-type extension doesn't work nowadays, so must use default file format (png).

File-type extension examples:
* good url (without file ext): https://www.gravatar.com/avatar/a443dd413b286902c47da02fe0a56e5e?d=404&s=80
* broken url (with file ext): https://www.gravatar.com/avatar/a443dd413b286902c47da02fe0a56e5e.jpg?d=404&s=80


umputun (approved):
lgtm, thx

PR #167, opened 2023-06-16
site & sess legacy, for back compat

sowiner:
legacy, for back compat


umputun (changes requested):
I think both aud and sess code and be deduplicated by adding a pair of functions to provider, i.e. getAud(r *http.Request) and similar getSess

PR #184, opened 2023-10-24
Remove bluemonday and fix double-escaping

### LINKED ISSUE #183: remove bluemonday as an unreliable dependency
It seems the "bluemonday" dependency (used for HTML sanitisation) has a policy of unilaterally deleting their old versions shortly after releasing a new version:

> [Version: v1.0.26](https://pkg.go.dev/github.com/microcosm-cc/bluemonday?tab=versions) Latest 
Published: Oct 12, 2023

As of today (Oct 16), we see:

> go: warning: github.com/microcosm-cc/bluemonday@v1.0.25: retracted by module author: Retract older versions as only latest is to be depended upon
> go: to switch to the latest unretracted version, run:
> 	go get github.com/microcosm-cc/bluemonday@latest

We have also seen the same happen on the 22nd of July 2022.

This is potentially very problematic, as it means old versions of applications cannot be rebuilt.

Since this dependency is only being used in a couple of locations, it shouldn't be difficult to remove it entirely, sidestepping the version management issue.

The locations it's currently used:

1. [verify.sanitise](https://github.com/go-pkgz/auth/blob/f2975ebffaa3731be15414aead54479aa560d39b/provider/verify.go#L215C24-L215C32)
2. [custom_server](https://github.com/go-pkgz/auth/blob/f2975ebffaa3731be15414aead54479aa560d39b/provider/custom_server.go#L163)

Both of these seem to be incorrect uses of HTML sanitisation anyway: the data is provided to JSON structures (the second one is being built manually and does need some escaping; currently it would actually break / be exploitable if the user had a `"` or `\` in their user ID, but that's perhaps a separate issue). The server doesn't need to strip HTML tags in the first place, as handling special characters is the concern of whatever later displays this data on a page (and should be achieved by escaping rather than stripping).

david-bezero:
Fixes #183

Uses `json.Marshal` to build a correctly encoded user structure (replacing `fmt.Sprintf`), removing bluemonday from this part of the code entirely as it is not required.

Also updates the email template code to avoid double-encoding everything, and removes bluemonday as it is not necessary (any tags entered are encoded by the `template` library anyway, as seen in the test, so there is no risk of injection)

This preserves the whitespace / trimming tidy-up steps for the email data.


provider/custom_server.go
```diff
@@ -160,16 +160,19 @@ func (c *CustomServer) handleUserInfo(w http.ResponseWriter, r *http.Request) {
...
 	userID := ti.GetUserID()
 
-	p := bluemonday.UGCPolicy()
-	ava := p.Sanitize(fmt.Sprintf(c.URL+"/avatar?user=%s", userID))
-	res := fmt.Sprintf(`{
-					"id": "%s",
-					"name":"%s",
-					"picture":"%s"
-					}`, userID, userID, ava)
+	user := token.User{
```
umputun, line 163: I don't understnd how json marshal prevents injection of userID here. Even if it does somehow, it is missing a test demonstrating it.

provider/verify.go
```diff
@@ -134,9 +133,7 @@ func (e VerifyHandler) LoginHandler(w http.ResponseWriter, r *http.Request) {
 // GET /login?site=site&user=name&address=someone@example.com
 func (e VerifyHandler) sendConfirmation(w http.ResponseWriter, r *http.Request) {
 
-	user, address := r.URL.Query().Get("user"), r.URL.Query().Get("address")
-	user = e.sanitize(user)
-	address = e.sanitize(address)
+	user, address, site := r.URL.Query().Get("user"), r.URL.Query().Get("address"), r.URL.Query().Get("site")
```
umputun, line 136: this one is unclear to me either. The PR comment says "bluemonday as it is not necessary (any tags entered are encoded by the template library anyway, so there is no risk of injection)" but this method is a handler, and nothing prevents bad actors to pass anything in.

provider/verify_test.go
```diff
@@ -77,20 +78,22 @@ func TestVerifyHandler_LoginSendConfirmRejected(t *testing.T) {
...
-	req, err := http.NewRequest("GET", "/login?address=blah@user.com&user="+badUser+"&site=remark42", http.NoBody)
+	badData := "<html><script>nasty stuff</script>&lt;escaped&gt;</html>"
```
umputun, line 81: I don't get why this test changed this way and the user is url-escaped as a part of the test. It looks like you don't expect badUser to be passed in, but there is nothing preventing it.

umputun (changes requested):

provider/custom_server.go
david-bezero: it doesn't prevent html tags - it prevents backslashes and quotes (backslashes weren't caught with the old approach, so any ID ending with a backslash would break the JSON syntax)

the point is that HTML tags don't _need_ to be escaped here: none of this stuff ends up as HTML (the picture might be rendered in a `src` attribute, but then it's up to the consumer to escape it at that point. It could just as easily be passed in to a `fetch` or similar, which would need different escaping semantics, so escaping it here is the wrong location.

for tests, this is covered by the existing tests, but a new one could be added with a backslash in the ID to prevent future regressions.

provider/verify.go
david-bezero: the `sanitize`'d values are passed to 2 places:

- a handshake (which doesn't need them to be escaped or sanitised because it just handles them as a string)
- an email template builder (which does its own escaping)

provider/verify_test.go
david-bezero: I mostly changed this to make it easier to reason about what's going on. The old test value was _very_ long, and had its own hard-coded URL encoding, so to keep the test simple I replaced it with a shorter demonstration of the attack it's trying to prevent, and used the built-in function to automatically URL-encode it.

I renamed it from `badUser` to `badData` because I added it to the `site` parameter as well (which previously wasn't being tested for how it handles malicious data)

david-bezero:
I'm not sure what to make of the linting error but it's not related to anything that changed in this PR:

> G101: Potential hardcoded credentials: RSA private key (gosec)

(file: provider/apple_test.go)

provider/verify.go
umputun: @paskal - can you take a look pls? To me, it looks like the `Sender` called after html/template was applied, and it can be safe to avoid sanitizing here. However, as far as I recall, this code was added to address some potential vulnerabilities, and I don't recall many details about it anymore. 

This sanitization was added by me on 06/03/22 "sanitize both user and address in verified provider". Do you recall if it was the result of some finding in remark42 around the same time?

provider/verify_test.go
umputun: I am almost sure this long and ugly input resulted from some vulnerability report we tried to replicate and address. I have asked @paskal to check it out and cross-reference, I will take a closer look too.

provider/custom_server.go
umputun: This sounds scary and means we are introducing a real attack vector to all the existing users who will have to handle potential injection via avatar's URL on some level. Avatars are here for only one reason - displaying them on web pages, and it is our responsibility to prevent possible attacks here.

david-bezero: a possible option which preserves the "safety net" for users who forget to escape things in HTML themselves, without compromising functionality for users who do the right thing, would be to URL-encode the value (excluding `/` but being sure to include `"`, `'`, `<`, and `>`). This will prevent any injection attacks if rendered as a raw string into HTML, without breaking use-cases where the value is used in non-html-source ways (e.g. passing to a `.src` attribute via javascript, or using `fetch`, or using a server-side request for the image)

I'd personally argue that this safety net isn't necessary, as users who forget to escape the value will almost certainly have plenty of other vulnerabilities in their own code anyway, but it's available as an option which maintains correctness.

provider/verify_test.go
david-bezero: The previous "badUser" input is nothing more than a URL encoded copy of a phishing message: if it is allowed to render raw, the user sees the phishing content before the real content of the message, and is encouraged to enter their password into an input field (in a real attack this would presumably submit somewhere).

As such, it's more of a demonstration of how this vulnerability could be used in a meaningful attack, than a carefully crafted string which triggers a vulnerability in itself. For the purpose of a test, I'd argue that keeping things simple is better. The much shorter input I swapped it with still tests the same attack vector, just without a full demonstration of the impact.

provider/custom_server.go
david-bezero: considering this further: since the user ID is being passed in as a query parameter in the avatar URL, it should be URL encoded anyway just for correctness. And since `c.URL` is controlled by the developer, that part of the URL does not carry any risk.

I'll add URL encoding for the userID parameter in this URL (which was missing before) to make it more correct, which will have a side-effect of making it safe even for users who forget to escape the string when injecting into raw HTML. I'll also add a test for this situation.

david-bezero: I have added `url.QueryEscape` around the userID and added a test to confirm it handles the important special characters.

provider/verify.go
david-bezero: I did a little digging in to how these checks came about and found your comment @umputun here:

https://github.com/go-pkgz/auth/pull/119#issuecomment-1146404853

That explains why you were previously vulnerable to this attack, but are now double-encoding. The `text` -> `html` template change was the only required change in the end - it made the rest of the changes obsolete, and this PR effectively undoes those parts.

paskal:
@david-bezero, please rebase from the master to fix tests. Are there any changes expected in this PR based on previous comments, or it's in the final state for the review?

paskal:
@david-bezero, gentle ping for an answer.

umputun:
Could you pls resolve the conflict? I'm going to merge it in and release a new version as soon as it done

david-bezero:
Sorry for the delay; I've been away over the holidays. Rebased and force-pushed.

umputun (approved):

PR #201, opened 2024-06-03
Fix registration of dev provider in Service.authMiddleware.Providers

cyb3r4nt:
There was one minor problem related to the dev and Apple providers registration.

Now it is possible to have a configuration,
where only one single dev provider is enabled.

Providers were not registered into `Service.authMiddleware.Providers` slice in the `Service.AddDevProvider()` and `Service.AddAppleProvider()` methods before.

This worked before, because other providers were registered at the same time, and `authMiddleware.Providers` slice was reassigned by other methods.

### DIFF
--- a/v2/auth_test.go
+++ b/v2/auth_test.go
@@ -227,7 +227,11 @@ func TestIntegrationAvatar(t *testing.T) {
 }
 
 func TestIntegrationList(t *testing.T) {
-	_, teardown := prepService(t)
+	_, teardown := prepService(t, func(svc *Service) {
+		svc.AddProvider("github", "cid", "csec")
+		// add go-oauth2/oauth2 provider
+		svc.AddCustomProvider("custom123", Client{"cid", "csecret"}, provider.CustomHandlerOpt{})
+	})
 	defer teardown()
 
 	resp, err := http.Get("http://127.0.0.1:8089/auth/list")
@@ -237,7 +241,7 @@ func TestIntegrationList(t *testing.T) {
 
 	b, err := io.ReadAll(resp.Body)
 	require.NoError(t, err)
-	assert.Equal(t, `["dev","github","custom123","direct","direct_custom","email"]`+"\n", string(b))
+	assert.Equal(t, `["dev","github","custom123"]`+"\n", string(b))
 }
 
 func TestIntegrationUserInfo(t *testing.T) {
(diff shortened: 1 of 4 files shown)


v2/auth_test.go
```diff
@@ -374,19 +382,28 @@ func TestDirectProvider(t *testing.T) {
 }
 
 func TestDirectProvider_WithCustomUserIDFunc(t *testing.T) {
-	_, teardown := prepService(t)
+	_, teardown := prepService(t, func(svc *Service) {
+		svc.AddDirectProviderWithUserIDFunc("directCustom",
```
cyb3r4nt, line 386: Note that using a provider name with `_`  symbol breaks some expectations in [Authenticator.isProviderAllowed()](https://github.com/go-pkgz/auth/blob/master/v2/middleware/auth.go#L156-L167) method.

paskal:
I reviewed the changes and I think they are not breaking anything.

The only thing I still do not understand is the original issue, could you please describe it in more detail, and maybe provide a code snippet which reproduces it? Ideally a test which reproduces the problem in master.

cyb3r4nt:
> The only thing I still do not understand is the original issue, could you please describe it in more detail, and maybe provide a code snippet which reproduces it? Ideally a test which reproduces the problem in master.

I was trying to use this auth lib in my dev environment, and only a single dev provider was configured there.
It is possible to reproduce it with [`TestIntegrationProtected`](https://github.com/go-pkgz/auth/blob/v1.23.0/auth_test.go#L145) and commenting out [all other providers in prepService() preparation method](https://github.com/go-pkgz/auth/blob/v1.23.0/auth_test.go#L513-L533)
[This check](https://github.com/go-pkgz/auth/blob/v1.23.0/middleware/auth.go#L131-L136) was failing for me during the request to some protected web endpoint.
Same thing may be reproduced if [remark42](https://github.com/umputun/remark42/blob/v1.13.0/backend/app/cmd/server.go#L949-L957) is started only with option `--auth.dev` and without specifying any other providers.

The new and modified version of `TestIntegrationProtected` from this PR should cover this issue.

v2/auth_test.go
cyb3r4nt: I will add these changes to v1 as well.

cyb3r4nt: I copied same changes to `v1/auth.go` and `v1/auth_test.go` as well.
These files are same in v1 and v2 at the moment.

paskal (approved):

umputun (approved):
LGTM, thx

PR #203, opened 2024-06-17
Set proper "content-type" while calling JwtService.Reset()

iRay:
During "LogOut", JwtService.Reset() doesn't set "content-type" header that leads to "XML Parsing Error: no root element found" in FireFox.
This PR simply adds proper header.

### DIFF
--- a/auth_test.go
+++ b/auth_test.go
@@ -292,6 +292,7 @@ func TestLogout(t *testing.T) {
 	resp, err = client.Get("http://127.0.0.1:8089/auth/logout")
 	require.Nil(t, err)
 	assert.Equal(t, 200, resp.StatusCode)
+	assert.Equal(t, "application/json; charset=utf-8", resp.Header.Get("Content-Type"))
 	defer resp.Body.Close()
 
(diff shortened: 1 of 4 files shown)


paskal (approved):

paskal:
@iRay, please copy the exact change to the v2 directory, as it will be released soon and will need those changes.

iRay:
@paskal ok, got you, thx.

paskal:
@iRay apologies for not being clear on that: could you please add the change both to v1 and v2?

iRay:
@paskal you was clear enough and I added  the same changes to v2. Maybe I missed something else?

iRay:
> @paskal you was clear enough and I added the same changes to v2. Maybe I missed something else?

Or maybe you meant I need to create another branch and new PR with changes for v1 and v2?

umputun (approved):
LGTM, thx

paskal:
Now it's good, thank you!

paskal:
This commit caused issue https://github.com/umputun/remark42/issues/1812. I think changing the context type to `text/plain` would fix the situation.

PR #207, opened 2024-07-20
Implement XSRFIgnoreMethods

oalexander6:
# XSRFIgnoreMethods

## Background
I discovered while using this library that the current XSRF protections do not allow for a GET request sent by the browser on protected routes, due to requiring the presence of the XSRF token in a header. This is a problem for applications that are not SPAs and send GET requests to the server to retrieve a new HTML page. In these situations, the common pattern is to skip XSRF checks on GET requests.

## Solution
To resolve this, I implemented a configuration option called `XSRFIgnoreMethods`, allowing the user to specify a comma-separated list of HTTP methods for which XSRF checks should be skipped when extracting and validating the JWT. I am open to any feedback, modifications, or  discussions about this.

### DIFF
--- a/token/jwt.go
+++ b/token/jwt.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"slices"
 	"strings"
 	"time"
 
@@ -35,12 +36,11 @@ type Handshake struct {
 
 const (
 	// default names for cookies and headers
-	defaultJWTCookieName     = "JWT"
-	defaultJWTCookieDomain   = ""
-	defaultJWTHeaderKey      = "X-JWT"
-	defaultXSRFCookieName    = "XSRF-TOKEN"
-	defaultXSRFHeaderKey     = "X-XSRF-TOKEN"
-	defaultXSRFIgnoreMethods = ""
+	defaultJWTCookieName   = "JWT"
+	defaultJWTCookieDomain = ""
+	defaultJWTHeaderKey    = "X-JWT"
+	defaultXSRFCookieName  = "XSRF-TOKEN"
+	defaultXSRFHeaderKey   = "X-XSRF-TOKEN"
 
 	defaultIssuer = "go-pkgz/auth"
 
@@ -48,6 +48,10 @@ const (
 	defaultCookieDuration = time.Hour * 24 * 31
 
 	defaultTokenQuery = "token"
+)
+
+var (
+	defaultXSRFIgnoreMethods = []string{}
 )
 
 // Opts holds constructor params
@@ -65,7 +69,7 @@ type Opts struct {
 	JWTHeaderKey      string
 	XSRFCookieName    string
 	XSRFHeaderKey     string
-	XSRFIgnoreMethods string
+	XSRFIgnoreMethods []string
 	JWTQuery          string
 	AudienceReader    Audience      // allowed aud values
 	Issuer            string        // optional value for iss claim, usually application name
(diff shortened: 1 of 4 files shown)


umputun:
isn't this already exist with https://github.com/go-pkgz/auth/blob/master/auth.go#L106 https://github.com/go-pkgz/auth/blob/master/auth.go#L106 ?

oalexander6:
No, in this situation I still definitely want XSRF protections on POST, PUT, DELETE, etc. but specifically not for GET requests. The DisableXSRF option disables it for all methods.

oalexander6:
This PR would allow for ignoring just certain HTTP methods, by setting `XSRFIgnoreMethods: []string{"GET", "HEAD"}` for example, to just skip the XSRF checks for "GET" and "HEAD" only

token/jwt.go
```diff
@@ -59,17 +60,18 @@ type Opts struct {
...
-	Issuer          string        // optional value for iss claim, usually application name
-	AudSecrets      bool          // uses different secret for differed auds. important: adds pre-parsing of unverified token
-	SendJWTHeader   bool          // if enabled send JWT as a header instead of cookie
-	SameSite        http.SameSite // define a cookie attribute making it impossible for the browser to send this cookie cross-site
+	JWTCookieName     string
+	JWTCookieDomain   string
+	JWTHeaderKey      string
+	XSRFCookieName    string
+	XSRFHeaderKey     string
+	XSRFIgnoreMethods string
```
umputun, line 68: I prefer not to use a comma-separated string in this context and then apply `strings.contains` on the string subsequently. It would be more organized to utilize a slice of methods in this scenario and perform an exact match on the request method.

umputun:
> This PR would allow for ignoring just certain HTTP methods, by setting `XSRFIgnoreMethods: []string{"GET", "HEAD"}` for example, to just skip the XSRF checks for "GET" and "HEAD" only

Unless I missed something, this code does not expect a []string with all the ignored methods but rather a single string (see my comment in the review above).

token/jwt.go
oalexander6: Agreed, I originally did that but saw the the default values for the related options were constants and wanted to stick to that, which couldn't be done with a slice. I will switch it back to an array of strings.

umputun:
one more thing - this change should be also documented in readme. Maybe a section about "Options to ignore XSRF protections" or smth like what showing how to do it and providing a basic explanation why this may be needed

oalexander6:
> one more thing - this change should be also documented in readme. Maybe a section about "Options to ignore XSRF protections" or smth like what showing how to do it and providing a basic explanation why this may be needed

Will do

oalexander6:
The requested changes have been made. Thanks for the feedback.

token/jwt.go
```diff
@@ -293,7 +303,7 @@ func (j *Service) Get(r *http.Request) (Claims, string, error) {
 		return Claims{}, "", fmt.Errorf("token expired")
 	}
 
-	if j.DisableXSRF {
+	if j.DisableXSRF || slices.Contains[[]string, string](j.XSRFIgnoreMethods, r.Method) {
```
umputun, line 306: I don't think you need to implicitly specify types here as they should be inferred just fine. I believe `slices.Contains(j.XSRFIgnoreMethods, r.Method)` will work.

umputun (commented):
a minor thing with implicit types for Contains func

token/jwt.go
oalexander6: Made this change. Thanks

umputun (approved):
LGTM, thx

paskal:
@oalexander6  could you please copy the very same changes to v2 directory please?

oalexander6:
> @oalexander6 could you please copy the very same changes to v2 directory please?

https://github.com/go-pkgz/auth/pull/210

PR #209, opened 2024-07-29
Update go modules, youmark/pkcs8

paskal:
4 years of improvements arrived.


paskal:
I have no idea why test started failing but I can't reproduce it reliably enough locally to fix it. Something is off in the CI.

paskal:
Now it's good to go, I found where `testing.T` was used after the test termination.

PR #211, opened 2024-08-02
Add custom HTTP handler for authentication errors

cyb3r4nt:
This patch gives a possibility to configure custom HTTP handlers, which can change the default authentication error responses in the middlewares.

The default implementation uses `http.Error()` method, which produces content type `text/plain`
 and simple textual messages rendered directly into the response body.
 In some applications this behavior may not be enough,
and there may be a need to produce different responses like JSON or HTML.
The `AuthErrorHttpHandler` interface allows to intercept authentication errors and write HTTP responses according to specific application requirements.

I also added some additional middleware methods, which allow to configure different `AuthErrorHttpHandler` implementations for different HTTP routes. Some routes in the applications may require different response bodies or attributes.


README.md
```diff
@@ -359,8 +359,9 @@ There are several ways to adjust functionality of the library:
...
 1. `Validator` - interface with `Validate(token string, claims Claims) bool` method. This is post-token hook and will be called on **each request** wrapped with `Auth` middleware. This will be the place for special logic to reject some tokens or users.
 1. `UserUpdater` - interface with `Update(claims token.User) token.User` method.  This method will be called on **each request** wrapped with `UpdateUser` middleware. This will be the place for special logic modify User Info in request context. [Example of usage.](https://github.com/go-pkgz/auth/blob/19c1b6d26608494955a4480f8f6165af85b1deab/_example/main.go#L189)
+1. `AuthErrorHTTPHandler` - interface with `ServeAuthError(w http.ResponseWriter, r *http.Request, and other params)` method. It is possible to change how authentication errors are written into HTTP responses by configuring custom implementations of this interface for the middlewares.
 
-All of the interfaces above have corresponding Func adapters - `SecretFunc`, `ClaimsUpdFunc`, `ValidatorFunc` and `UserUpdFunc`.
+Some of the interfaces above have corresponding Func adapters - `SecretFunc`, `ClaimsUpdFunc`, `ValidatorFunc` and `UserUpdFunc`.
```
paskal, line 364: "All except `AuthErrorHTTPHandler`" would be better.

auth_test.go
```diff
@@ -246,6 +247,206 @@ func TestIntegrationList(t *testing.T) {
...
+	reason string,
+	statusCode int,
+) {
+	h.wasCalled = true
+	w.Header().Set("Content-Type", h.contentType)
+	w.WriteHeader(h.statusCode)
+	fmt.Fprint(w, h.responseBody)
+}
+
+func TestIntegrationAuthErrorHTTPHandler(t *testing.T) {
```
paskal, line 270: This would be way more compact and readable if done as a table test. E.g. you need to create a table of `testAuthErrorHTTPHandler` and expected outcomes and test them in a loop with same setup and same checks.

paskal (approved):
Overall looks good, just two minor nitpicks.

README.md
cyb3r4nt: Changed

auth_test.go
cyb3r4nt: I have removed some duplication from this test and described different API calls as a table.

paskal (approved):

umputun:
> "In some applications, this behavior may not be enough, and there may be a need to produce different responses like JSON or HTML." 

I don't really get it. What are those cases that need a custom body in response, in addition to the status code? So far, I have seen one semi-convincing case where UI code expected some JSON body in all responses, including 401/403, because it had some trouble handling content-type. To me, it is not a good reason for the change, but in case you have some valid use cases, please share.

cyb3r4nt:
I am trying to create an application where server side page rendering is used (go `html/template`, [templ](https://github.com/a-h/templ) or any other template engine).
If some protected page is opened in the browser without required authentication or authorization,
then i want to display a proper error page to the user,
and pick some template based on the auth outcome.
It is a simple exchange of the HTML documents without any AJAX support in the client.

This scenario is possible because there is an HttpOnly JWT cookie,
which is sent by browser in HTTP requests.

Right now it is not possible to output a custom response to user
because auth middleware works before any application logic,
and auth middleware writes its own HTTP response to the `http.ResponseWriter`.

It may be possible to wrap auth middleware and pass a fake `http.ResponseWriter` to the underlying logic, 
including auth middleware itself,
and then decide what to do after wrapped middleware chain completes,
but such use case will require unnecessary copying for valid and authenticated requests,
which count is usually greater than count of auth failures.

In addition, i want to replace the `logger.L` with my own logger with structured logging and tracing support.

Such proposed `AuthErrorHTTPHandler` allows to control what is written to HTTP response,
and execute some application specific logic when authentication/authorization has failed.

This is what i want to achieve by introducing this change.

umputun:
> Right now it is not possible to output a custom response to user because auth middleware works before any application logic, and auth middleware writes its own HTTP response to the http.ResponseWriter.

I don't get this part. There is nothing in the auth middleware forcing you to put it first in the chain of middlewares. You can have others, like logging, panic/recovery, and the custom 4xx status code handling before the auth middleware in your chain.

Unless I'm missing something, I think you can just add your own middleware prior to the auth middleware in your chain and implement your own handling of status codes with a custom implementation of the ResponseWriter. I don't see why it would need to care about the underlying auth middleware, as all you do here is inspect the status code set by auth (or any other middleware in the chain it is wrapping).

_Generally speaking, to me, the auth library seems to be the wrong place to provide rendering-related customizations._

cugu:
I'd love to see this merged. I am working on a SCIM server, which requires a [specific format](https://datatracker.ietf.org/doc/html/rfc7644#section-3.12) for error messages.

Writing a custom ResponseWriter as suggested above gives me less control over the details. If I get a 401 in my custom ResponseWriter how do I know if it is from `go-pkgz/auth` or somewhere further down the chain? I'd probably also need to buffer and inspect the written body data, before writing it back to the client.

PR #213, opened 2024-08-29
Improve provider name handling

cyb3r4nt:
The problem occurs when the provider name contains underscore characters `_`.
If provider name is like `provider_prod` and full `claims.User.ID`  in the JWT token looks like `provider_prod_user1`,
then [Authenticator.isProviderAllowed()](https://github.com/go-pkgz/auth/blob/master/v2/middleware/auth.go#L156-L167) check fails and provider with such name cannot be used.
This was initially discovered in  https://github.com/go-pkgz/auth/pull/201#discussion_r1624361335.

It might be better to add an explicit provider name into the JWT token claims,
and avoid parsing already serialized string back to tokens.

Provider name passed into `Service.AddProvider()`  also becomes a part of  `https://host:port/auth/provider_prod/login`  URL, and therefore it requires special handling. 
One solution is to url-encode it, but then it will be still possible to use names containing spaces or special characters (by accident or with purpose).
Another solution is to forbid all provider names which require url-encoding.
It might be better to forbid empty names as well.
`_` underscore has been mentioned  in the README examples for some time now, i am not sure about it.
But those names may be even more strict and contain only ASCII alphanumeric symbols.
What do you think?

It is not possible to return errors from `Service.AddProvider()`, therefore invalid providers are just ignored and `ERROR` level message is logged.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -4,6 +4,8 @@ package auth
 import (
 	"fmt"
 	"net/http"
+	"net/url"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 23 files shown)


cyb3r4nt:
golangci-lint fixes are in https://github.com/go-pkgz/auth/pull/214

paskal (approved):
LGTM, I guess it shouldn't break anything.

I would suggest creating separate small PR with all changes not related to name handing improvement, e.g. README.md, provider/apple_pubkeys.go, provider/apple_test.go, provider/telegram_test.go.

paskal:
Rebased on top of master, without altering anything. @umputun could you please take a look?

auth.go
```diff
@@ -270,10 +272,42 @@ func (s *Service) addProviderByName(name string, p provider.Params) {
...
+		return false
+	}
+
+	formatForbidden := func(name string) {
+		s.logger.Logf("[ERROR] provider has been ignored because its name contains forbidden characters: '%s'", name)
+	}
+
+	path, err := url.PathUnescape(name)
+	if err != nil || path != name {
+		formatForbidden(name)
```
umputun, line 294: small thing: in debug mode, this function's redirect for logging will lose the source line information. I'd suggest creating an error message instead and logging it directly.

umputun (commented):
thx, lgtm. just a minor suggestion

auth.go
paskal: Rebased, addressed in the last commit.

PR #230, opened 2025-02-03
feature: Twitch OAuth2 Provider

pixel365:
Added support for authentication via Twitch

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@
 .vscode
 *.cov
 Dockerfile
+.idea
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # auth - authentication via oauth2, direct and email
 [![Build Status](https://github.com/go-pkgz/auth/workflows/build/badge.svg)](https://github.com/go-pkgz/auth/actions) [![Coverage Status](https://coveralls.io/repos/github/go-pkgz/auth/badge.svg?branch=master)](https://coveralls.io/github/go-pkgz/auth?branch=master) [![godoc](https://godoc.org/github.com/go-pkgz/auth?status.svg)](https://pkg.go.dev/github.com/go-pkgz/auth?tab=doc)
 
-This library provides "social login" with Github, Google, Facebook, Microsoft, Twitter, Yandex, Battle.net, Apple, Patreon, Discord and Telegram as well as custom auth providers and email verification.
+This library provides "social login" with Github, Google, Facebook, Microsoft, Twitter, Yandex, Battle.net, Apple, Patreon, Discord, Telegram and Twitch as well as custom auth providers and email verification.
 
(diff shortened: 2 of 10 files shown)


umputun:
twitch is a reasonable provider to have, but this implementation isn't the way, and after 17 months the branch conflicts with master. Closing this one, details below in case you want to open a fresh PR.

the main problem is that `TwitchHandler` re-implements the whole login/callback/logout flow instead of using `Oauth2Handler`. The only thing twitch needs that the generic handler can't do today is send a `Client-ID` header on the user info request, `oauth2.go:169` does a bare `client.Get(p.infoURL)`. The `{"data":[...]}` envelope is not a blocker, `mapUser` already gets the raw bytes and `NewPatreon`/`NewFacebook` unmarshal nested shapes that way.

what I'd do instead: add an optional `infoHeaders map[string]string` to `Oauth2Handler`, swap the `client.Get` for `NewRequestWithContext` + `client.Do` with those headers, and `NewTwitch` becomes a ~25 line entry in `providers.go` like every other provider. `apple.go` and `telegram.go` are hand written handlers, true, but they still implement the full contract. Apple reads `from` (`apple.go:262`), reads `session` into `SessionOnly` (`apple.go:264`) and validates the redirect host (`apple.go:409`).

as written the fork drops:

- `from`, never stored in the handshake and never redirected to, `AuthHandler` ends at `rest.RenderJSON` (`twitch.go:157`). User ends up on raw JSON instead of back where he started
- `SessionOnly`, hardcoded to `false` (`twitch.go:149`)
- `NoAva`
- `UserAttributes`, so `AddProviderWithUserAttributes` is a no-op for twitch
- `bearerTokenHook`

bugs:

1. `twitch.go:270`, `raw.Data[0]` with no bounds check. A valid `{"data":[]}` response parses fine and then panics
2. `twitch.go:269`, the unmarshal error is swallowed, so a malformed body gives back an empty `token.User` and auth continues with a blank identity
3. `twitch.go:241-246`, `Decode` can return before `defer res.Body.Close()` is registered, leaking the body on the error path. `requestUserInfo` gets this ordering right
4. `twitch.go:252`, status checked after decoding, so a non-200 with a non-JSON body surfaces as "unmarshalling data failed" and hides the real status
5. `twitch.go:265`, `scopes` is set and never read, `redirectURL` hardcodes the scope at `twitch.go:178`

and one that blocks on its own: `twitch.go:120` logs the whole token response, putting access and refresh tokens into the debug log. That is what 54502ee took out of the other providers.

smaller things: `.idea` in `.gitignore` is unrelated to this change, `AccessTokenResponse` doesn't need to be exported and its `Scope []string` is twitch specific (RFC 6749 has scope as a space delimited string), README advertises twitch while v1 and `_example` don't have it.

if you still want twitch, open a new PR against current master built on the `Oauth2Handler` extension and I'll review it.

PR #241, opened 2025-03-24
Set cookies for OAuth flows when SendJWTHeader is enabled

paskal:
This fixes https://github.com/umputun/remark42/issues/1877 where OAuth
 authentication fails when the send-jwt-header option is enabled. The problem occurred
 because:

1. When SendJWTHeader is enabled, the auth service only sends the JWT as a header
   without setting cookies
2. During OAuth flows, the authentication involves redirects between the app and the provider
3. HTTP headers don't persist through redirects, so the authentication state was lost

The solution:
- Modified the jwt.go token Set method to always set cookies during OAuth handshake phases (when claims.Handshake != nil), even when SendJWTHeader is enabled
- For normal authentication (non-handshake), maintain the original behavior where SendJWTHeader=true will only set headers
- This ensures the OAuth flow works properly while maintaining the correct behavior for API requests


paskal:
This diff doesn't work for Remark42 as-is because of reasons described in https://github.com/umputun/remark42/issues/1877#issuecomment-2840461165, but it would work oAuth for users of the library. It just seems that no one ever set this flag to hit the buggy behaviour with oAuth.

PR #245, opened 2025-04-29
Make all cookies HttpOnly for better security

paskal:
Adds HttpOnly flag to all cookies in Set and Reset functions to improve security by preventing JavaScript access. This protects cookies from XSS attacks.

In Reset function, the cookies are not read by JavaScript but are updated with HttpOnly for consistency with Set function behavior.


paskal:
Nvm, it doesn't make things more secure: https://security.stackexchange.com/a/229734

PR #273, opened 2026-04-11
Update dependencies to latest versions

paskal:
## Summary

- Update all direct and indirect dependencies to latest versions across v1, v2, and _example modules
- Fix v1 test compatibility: switch `custom_server_test.go` jwt import from v3 to v5 to match updated `go-oauth2/oauth2/v4` requirement
- Keep Go 1.24.0 in v2 and _example by pinning `golang.org/x` packages to latest 1.24-compatible versions

Notable dependency updates (v1 — Go 1.25.0):
- `golang.org/x/oauth2` v0.33.0 → v0.36.0
- `golang.org/x/image` v0.38.0 → v0.39.0
- `golang.org/x/crypto` v0.45.0 → v0.50.0
- `github.com/go-pkgz/rest` v1.20.4 → v1.21.0
- `github.com/go-pkgz/repeater/v2` v2.0.1 → v2.2.0
- `github.com/go-oauth2/oauth2/v4` v4.5.2 → v4.5.4
- `github.com/klauspost/compress` v1.18.2 → v1.18.5

Notable dependency updates (v2, _example — Go 1.24.0):
- `golang.org/x/crypto` v0.45.0 → v0.48.0
- `golang.org/x/image` v0.33.0 → v0.36.0
- `golang.org/x/oauth2` v0.33.0 → v0.35.0
- `golang.org/x/sync` v0.18.0 → v0.19.0
- `golang.org/x/sys` v0.38.0 → v0.41.0
- `golang.org/x/text` v0.31.0 → v0.34.0
- `github.com/go-pkgz/rest` v1.20.4 → v1.21.0
- `github.com/go-pkgz/repeater/v2` v2.0.1 → v2.2.0
- `github.com/golang-jwt/jwt/v5` v5.3.0 → v5.3.1
- `github.com/klauspost/compress` v1.18.2 → v1.18.5

PR #275, opened 2026-04-18
fix: validate "from" redirect target in OAuth/verify flows

paskal:
The `from` query parameter accepted by `oauth1`, `oauth2`, `apple` and `verify` login flows tells the server where to send the user after a successful auth handshake. Before this PR the value was stored verbatim in the handshake JWT and used as the redirect target with no validation — any external URL fits, so a crafted login link could route a user through legitimate provider OAuth and then bounce them to an attacker-controlled landing page.

## Attack

```
https://comments.example.com/auth/github/login?from=https://evil.example.com/phish
```

Victim sees the real GitHub consent screen and approves. Provider then calls back to the legitimate AuthHandler, which (pre-fix) executes:

```go
http.Redirect(w, r, oauthClaims.Handshake.From, http.StatusTemporaryRedirect)
```

The 307 lands the user on `evil.example.com/phish`, carrying a fresh session cookie and the trust the user had in the legitimate brand. `From` is JWT-signed so it cannot be tampered mid-flow — but the *initial* link is attacker-controlled and any URL fits inside the JWT.

The same pattern lives in all four providers: `oauth1.go:166`, `oauth2.go:242`, `apple.go:396`, `verify.go:130`.

## Fix

1. New `token.AllowedHosts` interface + `token.AllowedHostsFunc` adapter — mirrors the existing `Audience`/`AudienceFunc` pattern for consistency.
2. New `Opts.AllowedRedirectHosts` field threaded through `provider.Params` (and added directly to `VerifyHandler` since it doesn't embed `Params`).
3. Centralised `provider.isAllowedRedirect` helper — relative URLs and unparseable URLs are rejected; the host of `Opts.URL` is always permitted; additional hosts are opt-in via `AllowedRedirectHosts`.
4. All four redirect sites gate on the helper and fall back to the existing JSON user-info response on rejection (with a `[WARN]` log naming the disallowed host).

**Default behaviour (nil allowlist) accepts only the same host as `Opts.URL`** — safe for every existing single-host caller without any config change.

For multi-host deployments, callers register additional hosts:

```go
opts := auth.Opts{
    URL: "https://app.example.com",
    AllowedRedirectHosts: token.AllowedHostsFunc(func() ([]string, error) {
        return []string{"admin.example.com", "billing.example.com"}, nil
    }),
}
```

## Reproduction

`TestOauth2LoginFromRejectsExternalHost` in `provider/oauth2_test.go` exercises the full handler with `from=https://evil.example.com/phish` and asserts no external redirect is attempted (response is the JSON user info). The test docblock spells out the attack scenario for future readers. Same pattern repeated in `oauth1_test.go`, `apple_test.go`, `verify_test.go`. `TestOauth2LoginFromAllowsAllowlistedHost` covers the positive case (allowlisted host gets the 307).

`TestIsAllowedRedirect` in `provider/redirect_test.go` covers the validator with 14 cases: empty/relative/unparseable URLs, same-host with various scheme/port combinations, subdomain non-matching, allowlist hits and misses, getter errors.

## Verification

```
$ cd v2 && go test -race ./...
ok  github.com/go-pkgz/auth/v2          1.07s
ok  github.com/go-pkgz/auth/v2/avatar
ok  github.com/go-pkgz/auth/v2/logger
ok  github.com/go-pkgz/auth/v2/middleware
ok  github.com/go-pkgz/auth/v2/provider
ok  github.com/go-pkgz/auth/v2/provider/sender
ok  github.com/go-pkgz/auth/v2/token
```

`go vet` clean. New code passes `golangci-lint` (existing pre-PR findings in `telegram.go`, `providers.go`, `apple.go` are unchanged and unrelated).

## Files

- `v2/token/jwt.go` — new `AllowedHosts`/`AllowedHostsFunc`
- `v2/provider/redirect.go` (new) — `isAllowedRedirect` helper
- `v2/provider/redirect_test.go` (new) — 14 unit cases
- `v2/provider/oauth1.go` + `oauth1_test.go` — gate + reproduction test
- `v2/provider/oauth2.go` + `oauth2_test.go` — `Params.AllowedRedirectHosts` field, gate + 2 tests
- `v2/provider/apple.go` + `apple_test.go` — gate + reproduction test
- `v2/provider/verify.go` + `verify_test.go` — `URL`/`AllowedRedirectHosts` fields on `VerifyHandler`, gate + 2 tests
- `v2/auth.go` — wire `Opts.AllowedRedirectHosts` into all 6 `provider.Params` constructions and into `VerifyHandler`
- `README.md` — new "Allowed redirect hosts" section + notes on the two `from=` API entries

Reported via private security audit of remark42 (which vendors this library); confirmed by code inspection here and against the live reference deployment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ Generally, adding support of `auth` includes a few relatively simple steps:
 
 For the example above authentication handlers wired as `/auth` and provides:
 
-- `/auth/<provider>/login?site=<site_id>&from=<redirect_url>` - site_id used as `aud` claim for the token and can be processed by `SecretReader` to load/retrieve/define different secrets. redirect_url is the url to redirect after successful login. Only redirect targets whose host matches `Opts.URL` or appears in `Opts.AllowedRedirectHosts` are honoured; others fall back to the user-info JSON response. See "Allowed redirect hosts" below.
+- `/auth/<provider>/login?site=<site_id>&from=<redirect_url>` - site_id used as `aud` claim for the token and can be processed by `SecretReader` to load/retrieve/define different secrets. redirect_url is the url to redirect after successful login. When `Opts.AllowedRedirectHosts` is set, only targets whose host matches `Opts.URL` or appears in the allowlist are honoured; others fall back to the user-info JSON response. With the default (nil) allowlist any non-empty `from` is honoured — see "Allowed redirect hosts" below for the opt-in details.
 - `/avatar/<avatar_id>` - returns the avatar (image). Links to those pictures added into user info automatically, for details see "Avatar proxy"
 - `/auth/<provider>/logout` and `/auth/logout` - invalidate "session" by removing JWT cookie
 - `/auth/list` - gives a json list of active providers
@@ -291,7 +291,7 @@ used as `Sender`.
 
 The API for this provider:
 
- - `GET /auth/<name>/login?user=<user>&address=<address>&aud=<site_id>&from=<url>` - send confirmation request to user. The `from` URL is subject to the same allowlist check described in "Allowed redirect hosts".
+ - `GET /auth/<name>/login?user=<user>&address=<address>&aud=<site_id>&from=<url>` - send confirmation request to user. When host validation is enabled via `Opts.AllowedRedirectHosts`, the `from` URL is subject to the same allowlist check described in "Allowed redirect hosts"; otherwise any non-empty value is honoured.
  - `GET /auth/<name>/login?token=<conf.token>&sess=[1|0]` - authorize with confirmation token
 
 The provider acts like any other, i.e. will be registered as `/auth/email/login`.
--- a/README.md
+++ b/README.md
@@ -495,11 +495,11 @@ In order to allow `aud` support the list of allowed audiences should be passed i
 
 ### Allowed redirect hosts
 
-The `from` query parameter accepted by oauth2/oauth1/apple/verify login flows tells the server where to send the user after a successful auth handshake. The value is signed into the handshake JWT, so it can't be tampered with mid-flow — but a caller can put any URL into the initial login link. To prevent phishing campaigns from leveraging legitimate OAuth flows to land victims on attacker-controlled pages, the library validates the `from` host before issuing the redirect.
+The `from` query parameter accepted by oauth2/oauth1/apple/verify login flows tells the server where to send the user after a successful auth handshake. The value is signed into the handshake JWT, so it can't be tampered with mid-flow — but a caller can put any URL into the initial login link. Without restriction this becomes a phishing vector: an attacker crafts a login link that bounces the victim through legitimate OAuth and then to an attacker-controlled landing page.
 
-Default behaviour (no extra configuration): only `from` URLs whose host matches the host of `Opts.URL` are honoured. All other `from` values are silently dropped and the handler falls back to returning the user-info JSON instead of redirecting. This is safe for any single-host deployment.
+**Default behaviour is permissive (`Opts.AllowedRedirectHosts` is nil):** any non-empty `from` value is honoured. This preserves the behaviour of versions before the redirect validator existed, so a dependency bump never silently breaks an existing deployment. **Hardening is opt-in.**
 
-To accept additional hosts (multi-tenant deployments, separate admin/app subdomains, etc.), set `Opts.AllowedRedirectHosts` to anything implementing the `token.AllowedHosts` interface:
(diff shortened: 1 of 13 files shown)


paskal:
Pushed two fixup commits addressing review:

**`8fd76b0`** — Copilot inline items:
- `isAllowedRedirect` now compares `u.Hostname()` instead of `u.Host`, so https://app.example.com:443 and https://app.example.com match (3 new test cases cover :443, :80, and :8080).
- All four `[WARN] rejected redirect to disallowed host: ...` log lines now use a new `redirectHostForLog` helper that returns just the hostname; attacker-supplied paths/queries no longer leak into operator logs. Covered by `TestRedirectHostForLog`.

**`975bb1f`** — your call on the default:
You're right, a dep bump shouldn't silently restrict behaviour. Flipped the semantics: nil `AllowedRedirectHosts` is now permissive (legacy behaviour, no host check); the validator only runs when the consumer explicitly opts in by setting the field. Single-host deployments can enable the policy with `token.AllowedHostsFunc(func() ([]string, error) { return nil, nil })` — the `Opts.URL` host is always implicit when policy is on. Test table reorganised, README rewritten to make the opt-in nature clear.

CI is running.

umputun (commented):
nice fix. a few things before merge:

1. **new lint finding** — `v2/provider/oauth2_test.go:301:13` introduces a G101 gosec issue on the inline `oauth2.Endpoint{AuthURL:..., TokenURL:...}` string literals. Either `//nolint:gosec // G101: test URL, not a credential` or build the URLs via `fmt.Sprintf` like `prepOauth2Test` does.

2. **case-insensitive hostname compare** — `v2/provider/redirect.go:44` uses `h == fromHost` (byte-exact). DNS hostnames are case-insensitive, and `checkAuds` in this same repo already uses `strings.EqualFold` (`v2/token/jwt.go:384`). `App.Example.Com` in `from` would be rejected even if the allowlist has `app.example.com`. Recommend `strings.EqualFold` for consistency. Worth a test case alongside it.

3. **scheme allowlist** — Copilot's second point wasn't addressed. In practice `Hostname() == ""` filters `javascript:`/`data:`, so the attack surface is narrow, but explicit `u.Scheme == "http" || u.Scheme == "https"` closes it for defense-in-depth. Low priority if you'd rather leave it.

4. **typed-nil interface pitfall** — `var fn provider.AllowedHostsFunc; opts.AllowedRedirectHosts = fn` produces a non-nil interface wrapping a nil func, so `allowed == nil` is false and `allowed.Get()` panics. Rare-consumer-error, but a one-line guard in `isAllowedRedirect` (or just a godoc note on `AllowedHostsFunc`) would be safer.

question: is there a concrete dynamic-allowlist use case driving the `AllowedHosts` interface, or is this purely parity with `Audience`? `[]string` would be simpler if every caller passes a static list — but the parity argument is fine if you want it.

stale-base note: the README CrossOriginProtection section and go.mod downgrades in the diff are stale-base artifacts from #273/#274 landing on master — they'll disappear on rebase, not real changes.

umputun (commented):
all four items from my last pass landed cleanly — `strings.EqualFold`, explicit http/https scheme check, typed-nil `AllowedHostsFunc` guard, and the `prepOauth2Test`/`paramOpts` refactor that cuts ~60 lines and drops the G101. nice.

on Copilot's new six comments: they're all one issue — when you flipped semantics in `975bb1f` to opt-in, `redirect.go`'s docstring got updated but the downstream field godoc and README endpoint docs still describe the old "nil = same-host-only" behaviour. concretely:

- `v2/auth.go:68` — `Opts.AllowedRedirectHosts` says "Nil means same-host-only"
- `v2/provider/oauth2.go:47` — `Params.AllowedRedirectHosts` says "Nil disables additional hosts (same-host only)"
- `v2/provider/verify.go:36` — `VerifyHandler.AllowedRedirectHosts` says "Nil means same-host-only"
- `README.md:185` — reads as if validation is always on ("Only redirect targets whose host matches `Opts.URL` or appears in `Opts.AllowedRedirectHosts` are honoured")
- `README.md:294` — verify endpoint claims `from` is "subject to the same allowlist check" with no opt-in caveat
- `v2/provider/oauth2_test.go:268` — test docblock says "with the default (nil) allowlist" but the test wires non-nil via `enablePolicy`

all pure doc/comment fixes — one pass aligning the field godoc and README to the opt-in semantics that `redirect.go:13-16` already spells out correctly. then good to merge.

umputun (approved):
LGTM

PR #279, opened 2026-05-08
fix(middleware): don't log admin password on basic-auth failure

paskal:
## Summary

Admin basic-auth rejection log line currently includes the attempted password verbatim:

```
[WARN] admin basic auth failed, user/passwd mismatch, admin:hunter2
```

A mistyped legitimate password (often one character off the real one) or a misrouted password-manager submission lands in plaintext logs that may be shipped to centralised logging, crash bundles, or third-party observability systems.

## Change

Drop the password from the log line, keep the username:

```
[WARN] admin basic auth failed for user "admin"
```

Same fix in v1 (`middleware/auth.go:238`) and v2 (`v2/middleware/auth.go:246`), single PR.

## Test

New `TestBasicAdminUserDoesNotLogPassword` in both modules — captures the logger output, calls `basicAdminUser` with a wrong password, asserts the password substring is not in the captured output and the rejection is still logged.

Confirmed RED→GREEN locally:
- before: assertion fails, log contains `admin:super-secret-attempted-passwd`
- after: assertion passes

### DIFF
--- a/middleware/auth.go
+++ b/middleware/auth.go
@@ -235,7 +235,7 @@ func (a *Authenticator) basicAdminUser(r *http.Request) bool {
 
 	// using ConstantTimeCompare to avoid timing attack
 	if user != "admin" || subtle.ConstantTimeCompare([]byte(passwd), []byte(a.AdminPasswd)) != 1 {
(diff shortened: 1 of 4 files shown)


umputun (approved):
lgtm, thx

PR #284, opened 2026-05-08
fix(apple): redact tokens from exchange-response debug log

paskal:
## Summary

The Apple handler logged the full `appleVerificationResponse` struct on a `[DEBUG]` line:

```
[DEBUG] response data {AccessToken:M... TokenType:bearer ExpiresIn:3600
    RefreshToken:Iw... IDToken:eyJ...}
```

`AccessToken`, `RefreshToken` and `IDToken` are bearer credentials. With DEBUG-level logging enabled (default in many staging setups) these landed in stdout, file logs, centralized logging, crash bundles, and any third-party observability stack — anywhere log access doesn't imply auth-server-process compromise.

## Change

Replace the raw `%+v` dump with `appleVerificationResponseLogSummary`, which logs only the non-secret fields plus presence indicators (`present|missing`) for each token. Operators can still tell whether a response carried each token; the value never leaks.

```
[DEBUG] apple exchange response: type=bearer expires_in=3600
    access_token=present refresh_token=present id_token=present error=""
```

Same redaction in v1 (`provider/apple.go:334`) and v2 (`v2/provider/apple.go:334`), single PR.

## Test

`TestAppleVerificationResponseLogSummary` asserts the helper omits the three secret values verbatim and reports presence/missing correctly. Added in both modules. Full `go test -race ./...` green; `golangci-lint run --new-from-rev=master` 0 issues.

### DIFF
--- a/provider/apple.go
+++ b/provider/apple.go
@@ -331,7 +331,7 @@ func (ah AppleHandler) AuthHandler(w http.ResponseWriter, r *http.Request) {
 		rest.SendErrorJSON(w, r, ah.L, http.StatusInternalServerError, err, "exchange failed")
 		return
 	}
-	ah.Logf("[DEBUG] response data %+v", resp)
+	ah.Logf("[DEBUG] apple exchange response: %s", appleVerificationResponseLogSummary(resp))
 	if resp.Error != "" {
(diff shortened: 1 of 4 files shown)


umputun (approved):
lgtm, thx

PR #283, opened 2026-05-08
chore: go fix ./...

paskal:
## Summary

Apply Go modernization fixes from `go fix ./...` across both v1 and v2.

Mainly `interface{}` → `any` (the Go 1.18+ alias) plus small idiomatic updates the tool emits. Found while preparing PRs #279/#280/#281/#282 — kept separate so those security PRs stay focused.

## Change

- 36 files updated, ~84 insertions / ~86 deletions
- No behaviour changes; pure hygiene
- Go 1.25+ as required by `CLAUDE.md` and `go.mod`

## Test

`go test -race ./...` green in both v1 and v2.

PR #285, opened 2026-05-08
fix(sender/email): don't log email body, log size instead

paskal:
## Summary

The verify provider sends one-shot magic-link tokens via email. The Email sender's debug log dumped the full body verbatim:

```
[DEBUG] send "Confirmation token: <jwt>" to victim@example.com
```

Anyone with log access (centralized logging, crash bundles, support tools, mail-gateway observability) could redeem that token within its TTL — independently of the legitimate recipient. PR #281 (verify replay) limits reuse to one consumption, but the log reader can still race the user for that one consumption. This PR closes the leak at the source.

## Change

Log the recipient and body length only:

```
[DEBUG] send 142-byte message to victim@example.com
```

Same redaction in v1 (`provider/sender/email.go:84`) and v2 (`v2/provider/sender/email.go:84`), single PR.

## Test

`TestEmail_SendDoesNotLogBody` captures logger output, sends a known-secret body to a non-existent SMTP host (Send fails fast — that's fine, we only assert on the log line), and verifies:
- the body substring is absent from the log
- the recipient address is present (operators still need this for ops)

Added in both modules. Full `go test -race ./...` green; `golangci-lint run --new-from-rev=master` 0 issues.

### DIFF
--- a/provider/sender/email.go
+++ b/provider/sender/email.go
@@ -79,9 +79,13 @@ func NewEmailClient(emailParams EmailParams, l logger.L) *Email {
 	return &Email{EmailParams: emailParams, L: l, sender: sender}
 }
 
-// Send email with given text
+// Send email with given text. The body is not logged: confirmation emails
+// sent by the verify provider contain a one-shot magic-link token, and any
+// party with log access could redeem it before the user does. Logging only
+// the recipient and body length keeps the line useful for operators
+// without leaking the credential.
 func (e *Email) Send(to, text string) error {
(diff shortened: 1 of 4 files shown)


umputun (approved):
lgtm.

one small nit: the new test uses `func(format string, args ...interface{})` for the capturing logger, but #283 (just merged) modernized all existing `interface{}` to `any` across the codebase. Worth swapping the new test's signature to `args ...any` here too, otherwise it's a tiny regression on the modernization. Not blocking.

PR #287, opened 2026-05-08
fix(provider): bind dev oauth and custom-server to localhost by default

paskal:
## Summary

`provider/dev_provider.go` and `provider/custom_server.go` listened with `Addr: ":port"`, which binds to every interface. The dev OAuth UI and the embedded go-oauth2/oauth2 helper are intended for local development and embedded flows, not for LAN exposure -- but the bare-port shorthand silently exposed them to anyone who could route to the host's other addresses (other LAN clients, Docker network siblings, kubernetes pods on the same node, etc.).

## Change

Add `localBindAddr(host, port)` in `provider/service.go` that defaults the listen host to `127.0.0.1` when the caller passes an empty host. Callers that explicitly want non-loopback exposure can pass a hostname:

- **dev_provider** -- uses `Provider.Host` (existing field) as the host; empty -> `127.0.0.1`, explicit `"0.0.0.0"`/`"192.168.x.y"` honored.
- **custom_server** -- uses the host from `c.URL` (`http://localhost:8080`); `"localhost"` or empty -> `127.0.0.1`; explicit non-loopback honored.

Same change in v1 and v2 in single PR.

README updated with a "Listen address" section documenting:
- the security rationale for the localhost default;
- how to opt back in to all-interfaces binding (`AddDevProvider("0.0.0.0", port)` for dev; `URL: "http://0.0.0.0:port"` for custom server);
- the typical Docker case -- a server bound to `127.0.0.1` inside a container is not reachable from the host, so `0.0.0.0` is the right value when the auth process runs in a container.

Also fixed an outdated `AddDevProvider(port)` reference in the README -- the function takes `(host, port)`.

## Test

- `TestLocalBindAddr` -- table covering empty / explicit-loopback / explicit-LAN / explicit-`0.0.0.0` / IPv6.
- **Regression test:** `TestLocalBindAddr_DefaultIsNotAllInterfaces` -- asserts the default bind never produces the `":port"` shorthand. Reverting the helper to `return ":" + port` makes this test fail with: `Should not be: ":12345" -- default bind must not be all-interfaces`.

Both modules; `go test -race ./...` green; `golangci-lint run --new-from-rev=master` 0 issues.

## Migration note

Existing deployments that *intentionally* exposed the dev OAuth helper or custom server beyond loopback need to set `Provider.Host = "0.0.0.0"` (dev) or change their `Opts.URL` host to `0.0.0.0`/specific IP (custom-server). Default behaviour change is the security-relevant point of this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -547,7 +547,7 @@ Once the policy is on:
 
 ### Dev provider
 
-Working with oauth2 providers can be a pain, especially during development phase. A special, development-only provider `dev` can make it less painful. This one can be registered directly, i.e. `service.AddProvider("dev", "", "")` or `service.AddDevProvider(port)` and should be activated like this:
+Working with oauth2 providers can be a pain, especially during development phase. A special, development-only provider `dev` can make it less painful. This one can be registered directly, i.e. `service.AddProvider("dev", "", "")` or `service.AddDevProvider(host, port)` and should be activated like this:
 
 ```go
 	// runs dev oauth2 server on :8084 by default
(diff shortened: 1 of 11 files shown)


umputun (approved):
lgtm. Nice catch on the bundled deadlock fix in `CustomServer.Run`: pre-existing, `c.lock` was being held through the URL-parse and SplitHostPort error returns, which would block `Shutdown()`. The regression test `TestCustomServer_Run_BadURLDoesNotDeadlockShutdown` is well-shaped.

one tiny note: the `host == "localhost"` normalization in `custom_server.go:97` only normalizes the literal string. `127.0.0.1` and `[::1]` baked into `c.URL` would skip that branch, but they then go through `localBindAddr` → `net.JoinHostPort`, which produces a valid loopback bind. So no functional issue; just observing that the explicit "localhost" branch is needed only because `localBindAddr` treats empty as the default-loopback signal.

PR #288, opened 2026-05-09
Cap avatar fetch body size, pin redirect-validator bypass categories

paskal:
## Summary

Hardening followups from the security review of #275-#286 that don't fold into any of the four open security PRs.

### 1. `avatar.Proxy.Put` body-size cap (`avatar/avatar.go`, `v2/avatar/avatar.go`)

`Proxy.load` previously streamed the remote avatar `resp.Body` directly into `resize` without any cap. A malicious or buggy upstream sending an unbounded body could exhaust process memory.

This PR adds a 10 MiB cap (`maxAvatarFetchSize`):
- buffer the body through `io.ReadAll(io.LimitReader(resp.Body, maxAvatarFetchSize+1))` -- the +1 byte distinguishes "exactly at cap" from "too big".
- on overflow, return an error -- `Proxy.Put` then falls back to identicon as it does for any other load failure.
- legitimate avatars fit comfortably (Telegram caps photo at 5 MiB, Gravatar much smaller).

Test: `TestAvatar_PutCapsBodySize` simulates an upstream streaming ~12 MiB and asserts the user still gets a proxy URL (via identicon fallback) without OOM.

### 2. `isAllowedRedirect` bypass-category characterization tests (`v2/provider/redirect_test.go`)

Reviewer on #282 flagged untested bypass categories for the redirect validator. Verified manually that each is correctly rejected by the current `url.Parse` + `Hostname()` semantics; this PR pins that behaviour with a table of cases:

- scheme-relative `//evil.com/x`
- userinfo bypass `https://admin.example.com@evil.com/x`
- userinfo doesn't strip allowed host `https://evil@admin.example.com/x` (still allowed -- userinfo is irrelevant)
- IPv6 literal `https://[::1]/x` (rejected unless explicitly listed)
- IDN homoglyph (cyrillic `а`) and its punycode form not equal to ASCII allowlist entry
- percent-encoded host parse-fails
- backslash-userinfo trick parse-fails
- opaque `scheme:host` forms (`https:evil.com`, `https:///evil.com`)

Pure characterization tests -- no behaviour change, just guardrails so a future refactor doesn't silently weaken the validator.

## Items deferred to a separate followup once base PRs land

These touch code introduced by the open PRs (file doesn't yet exist on master), so they need to wait:

| Item | Depends on |
|---|---|
| `saveTelegramAvatar` body-size cap (mirror of the `Proxy.Put` cap above) | #286 |
| `botTokenInURLPath` regex anchoring (`/bot…/` vs `bot…` to stop matching `botFather`) | #286 |
| Rename `dumbAvatarSaver` test type to `legacyAvatarSaver` | #286 |
| Mirror these v2 bypass-category tests to v1 `provider/redirect_test.go` | #282 |
| v2 `validateAppleIDClaims` -> jwt v5 parser-options refactor (~70 lines drop) | #280 |

The HTTP client reuse nit on `saveTelegramAvatar` (#286 review item 6d) is intentionally skipped -- threading a shared `*http.Client` through the handler adds API surface for a one-call-per-login optimisation.

### DIFF
--- a/avatar/avatar.go
+++ b/avatar/avatar.go
@@ -26,6 +26,12 @@ import (
 // http.sniffLen is 512 bytes which is how much we need to read to detect content type
 const sniffLen = 512
 
+// maxAvatarFetchSize bounds the bytes read from a remote avatar URL. 10 MiB is
+// generous for any reasonable avatar (Telegram caps photo at 5 MiB; Gravatar is
+// much smaller); the cap protects Proxy.Put against an upstream sending an
+// unbounded body that would exhaust process memory inside resize.
+const maxAvatarFetchSize = 10 << 20
+
 // Proxy provides http handler for avatars from avatar.Store
 // On user login token will call Put and it will retrieve and save picture locally.
 type Proxy struct {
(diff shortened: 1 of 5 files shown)


paskal:
Folded the followups into the base PRs they belong to:

- avatar.Proxy.Put body-size cap (v1+v2) → moved into #286 (avatar safety)
- v2 isAllowedRedirect bypass-category tests → moved into #282 (also added v1 mirror)

Other deferred items also folded directly:
- saveTelegramAvatar body-size cap, regex anchoring, dumbAvatarSaver→legacyAvatarSaver rename → #286
- v2 jwt v5 parser-options refactor (drops validateAppleIDClaims helper) → #280

Nothing left for a standalone followup PR.

PR #289, opened 2026-05-09
fix(verify): close service-level typed-nil store + adapter-author guidance

paskal:
## Summary

Followups to #281 raised on the post-merge review. None blocking, all small.

| # | Item | Where |
|---|---|---|
| 1 | **Service-level typed-nil `VerifConfirmationStoreFunc` guard** -- silently disabled replay protection for `Opts{VerifConfirmationStore: VerifConfirmationStoreFunc(nil)}` | `auth.go`, `v2/auth.go` (+ tests) |
| 2 | `gofmt -w` after the `verifConfirmStoreO` -> `verifConfirmStoreOnce` rename | `auth.go`, `v2/auth.go` |
| 3 | Unit test for `scrubTokenFromRequest` -- the defensive early-return was uncovered after #281 | `provider/verify_test.go`, `v2/provider/verify_test.go` |
| 4 | Adapter-author godoc note on `MarkUsed` -- don't embed key in returned errors | `provider/verify.go`, `v2/provider/verify.go` |

## Item 1 in detail

The handler-level guard added in #281 round 2 (`provider/verify.go` LoginHandler) correctly normalises a typed-nil `VerifConfirmationStoreFunc` to no-store, but the Service-level check in `AddVerifProvider` was still:

```go
if s.opts.VerifConfirmationStore != nil {
    s.verifConfirmStore = s.opts.VerifConfirmationStore
    return
}
s.verifConfirmStore = provider.NewInMemoryVerifStore()
```

A typed-nil `VerifConfirmationStoreFunc` is a non-nil interface wrapping a nil func, so it survives the `!= nil` check, `NewInMemoryVerifStore()` is skipped, and at redemption time the handler-level guard then normalises the typed-nil to nil. **Net effect:** a user writing `Opts{VerifConfirmationStore: VerifConfirmationStoreFunc(nil)}` gets neither their func nor the default in-memory store, and replay protection is silently disabled.

Same shape as the `*avatar.Proxy` typed-nil case fixed in #286, with a different consequence (silent loss of protection rather than panic). The fix is the same shape applied one layer up:

```go
s.verifConfirmStoreOnce.Do(func() {
    store := s.opts.VerifConfirmationStore
    if fn, ok := store.(provider.VerifConfirmationStoreFunc); ok && fn == nil {
        store = nil
    }
    if store != nil {
        s.verifConfirmStore = store
        return
    }
    s.verifConfirmStore = provider.NewInMemoryVerifStore()
})
```

Regression test `TestService_AddVerifProvider_TypedNilStoreFuncFallsBackToDefault` in both v1 and v2: writes `var nilFn provider.VerifConfirmationStoreFunc; ConfirmationStore: nilFn`, asserts `verifConfirmStore` is non-nil and is **not** the typed-nil func itself (i.e. the default fired).

## Test

`go test -race ./...` green in both modules; `golangci-lint run --enable-only=misspell ./...` clean.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -27,15 +27,15 @@ type Client struct {
 
 // Service provides higher level wrapper allowing to construct everything and get back token middleware
 type Service struct {
-	logger             logger.L
-	opts               Opts
-	jwtService         *token.Service
-	providers          []provider.Service
-	authMiddleware     middleware.Authenticator
-	avatarProxy        *avatar.Proxy
-	issuer             string
-	useGravatar        bool
-	verifConfirmStore  provider.VerifConfirmationStore
+	logger                logger.L
+	opts                  Opts
+	jwtService            *token.Service
+	providers             []provider.Service
+	authMiddleware        middleware.Authenticator
+	avatarProxy           *avatar.Proxy
+	issuer                string
+	useGravatar           bool
+	verifConfirmStore     provider.VerifConfirmationStore
 	verifConfirmStoreOnce sync.Once
 }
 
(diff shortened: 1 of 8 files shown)


umputun (approved):
all 4 followups look good. Tests cover the typed-nil regression and `scrubTokenFromRequest`, v1/v2 mirror is right. LGTM.

PR #290, opened 2026-05-20
fix(avatar): prevent stored XSS via content-type spoofing

paskal:
## Summary

Fixes a stored XSS in `avatar.Proxy` and adds defense-in-depth headers across
the rest of `Service.Handlers()`. Both v1 (`avatar/`, `auth.go`) and v2
(`v2/avatar/`, `v2/auth.go`) on the same commit, so a single coordinated release
(`v1.x.y` and `v2.x.y` tags) covers both module versions.

## The bug

`avatar/avatar.go:Proxy.resize()` fell back to the raw bytes (`teeBuf`) when
`image.Decode` failed. `avatar/avatar.go:Proxy.Handler()` then called
`http.DetectContentType` on whatever was stored and emitted the result as
`Content-Type` — including `text/html` for HTML bodies.

### Attack scenario

The realistic path requires the attacker to control `u.Picture`:

1. Configure (or compromise) an OAuth provider so its `picture` claim points
   at attacker-controlled `https://evil.tld/avatar.png`.
2. Log in once. `Proxy.Put` fetches the URL, `resize` fails to decode the
   HTML body, falls through to `teeBuf`, stores the HTML under
   `<sha1(userID)>.image`.
3. Direct the victim to `https://auth-host/avatar/<sha1>.image`. Browser
   gets `Content-Type: text/html` with no `Content-Disposition`, renders the
   attacker's HTML as a document in the auth-service origin, runs script.

This is **not** exploitable on remark42's standard setup — `u.Picture` there
comes from admin-configured providers (GitHub, Google, etc.) which the user
cannot redirect. It **is** exploitable on:

- Custom OAuth providers whose `picture` claim is attacker-controlled.
- Self-signup flows that accept a Picture URL from the user.
- The dev provider (intentionally permissive, but still worth fixing).
- Any deployment where one user can set another's Picture (admin tooling, etc.).

Codex-flagged residual risks addressed in the same fix:

- **Decompression bomb** — a 100 KB compressed PNG declaring 65535×65535 px
  decodes to ~17 GB of pixel data. Before this PR `Proxy.resize` always called
  `image.Decode` and would OOM the auth service on a single login. After, an
  `image.DecodeConfig` pre-check rejects oversized declared dimensions before
  any pixel allocation.
- **32-bit overflow** — the pixel-count guard was computed in `int`, so on
  GOARCH=386 / 32-bit arm `cfg.Width * cfg.Height` for 16-bit GIF or JPEG
  dimensions could wrap below the cap. Multiplication is now in `int64`.
- **Animated GIF truncation** — the no-resize path now returns the original
  bytes verbatim; previously the `TeeReader` could capture only the first
  frame.

### Commit 1 — `fix(avatar): reject non-image content to prevent stored XSS via content-type spoofing`

`avatar/avatar.go` + `v2/avatar/avatar.go`:

- `Proxy.resize()` validates input via `image.DecodeConfig` (cheap, no pixel
  allocation) before any `image.Decode`. Refuses non-image content with `nil`
  so attacker payloads never reach `Store.Put`.
- New `maxAvatarPixels` constant (16 MP). Multiplication in `int64`.
- The no-resize path returns the original bytes verbatim.
- `Proxy.Put()` falls back to a generated identicon when `resize` returns `nil`.
- `Proxy.PutContent()` returns an explicit error for non-image input.
- `Proxy.Handler()` validates served bytes via the new `safeImgContentType`
  helper (allowlist: `image/{png,jpeg,gif,webp,bmp,x-icon}`; `image/svg+xml`
  explicitly rejected because SVG can execute scripts when navigated to top
  level). This catches stores poisoned before this fix.
- Every Handler response (success / 304 / error) carries:

  ```
  Content-Security-Policy: default-src 'none'; sandbox; frame-ancestors 'none'
  X-Content-Type-Options: nosniff
  Content-Disposition: inline; filename="avatar"
  ```

Tests updated: `TestAvatar_resize` and `TestAvatar_Routes` previously codified
the broken pass-through behavior; those assertions are inverted now.

New tests:

- `TestAvatar_PutRejectsNonImage` — store is not poisoned by attacker `u.Picture`.
- `TestAvatar_HandlerRejectsPoisonedStore` — handler refuses pre-poisoned bytes.
- `TestAvatar_resizeRejectsDecompressionBomb` — DecodeConfig + int64 dim cap.

`provider/telegram_test.go` placeholder `"png-bytes"` is now a real 1x1 PNG
(`tinyPNG`) since `PutContent` validates input.

### Commit 2 — `feat(auth): apply mandatory CSP and nosniff on every Service.Handlers() response`

`auth.go` + `v2/auth.go`:

`Service.Handlers()` wraps both returned handlers (auth route dispatcher and
avatar handler) with `withSecurityHeaders`. Sets:

```
Content-Security-Policy: default-src 'none'; sandbox; frame-ancestors 'none'
X-Content-Type-Options: nosniff
```

go-pkgz/auth's response surface is JSON-only for auth routes and images for
avatar — no consumer-customisable HTML anywhere — so this CSP is unconditionally
safe.

### Commit 3 — `refactor(avatar): collapse load/resize double-buffering`

`avatar/avatar.go` + `v2/avatar/avatar.go`:

Before this commit a remote avatar was buffered twice: once inside `load()` to
enforce `maxAvatarFetchSize`, again inside `resize()` for `image.DecodeConfig`.
For a 10 MiB upstream avatar each in-flight login held ~20 MiB peak.

- `load()` returns `([]byte, error)` instead of `(io.ReadCloser, error)`.
- `resize()` takes `([]byte, int)` instead of `(io.Reader, int)`.
- `PutContent()` now runs `io.ReadAll(io.LimitReader(content, cap+1))` at the
  public API boundary so the size cap is visible to callers reading the code.

No public API change, no behavior change other than halved peak memory per
concurrent login.

## What end-users will see

Anyone who upgrades and uses `Service.Handlers()`:

- Two new response headers on every auth and avatar response:
  - `Content-Security-Policy: default-src 'none'; sandbox; frame-ancestors 'none'`
  - `X-Content-Type-Options: nosniff`
- Avatar responses additionally carry `Content-Disposition: inline; filename="avatar"`.
- Non-image upstream avatars now fall back to a generated identicon instead
  of being stored verbatim. The user gets an identicon where previously they
  would have gotten a broken avatar (or, in the attacker case, JavaScript).
- `Proxy.PutContent` now returns an error for non-image input instead of
  silently storing whatever bytes the caller passed.

## What might break

Nothing for standard consumers. Specifically:

- **Embedders that proxy auth/avatar responses** — the CSP and nosniff
  headers are set on the response; downstream proxies pass them through.
  If a consumer was relying on the avatar handler returning `text/html` for
  invalid content, that was the bug — they should not be doing that.
- **Consumers calling `Proxy.PutContent` with invalid bytes** — they now
  get an error instead of an empty/poisoned avatar. This is the intended
  behavior.
- **SVG avatars** — explicitly rejected. SVG can execute scripts; storing
  one in an avatar slot was always a latent XSS risk.

## Release plan

Both `v1` and `v2` module versions live on `master` in this repo, so a single
PR fixes both. After merge, tag `v1.x.y` and `v2.x.y` from the same commit.
remark42 will bump to `v2.x.y` in its own follow-up PR.

## Codex review

The branch was reviewed adversarially with Codex (GPT-5.5, high reasoning) in
four rounds. Rounds 1 and 2 each surfaced one high-severity finding (the
decompression bomb on no-resize path, and the 32-bit int overflow on the
pixel-count guard) — both addressed in this PR. Rounds 3 and 4 approved with
no material findings.

## Tests

All v1 and v2 test suites pass, lint clean (no new `gosec`/`govet`/`misspell`
findings). Cross-compile to `GOARCH=386` succeeds, confirming the `int64`
arithmetic is portable.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -240,10 +240,11 @@ func (s *Service) Handlers() (authHandler, avatarHandler http.Handler) {
 }
 
 // withSecurityHeaders wraps an auth response handler to apply strict CSP and nosniff
-// on every response. The go-pkgz/auth package response surface is JSON-only for auth
-// routes and images for the avatar route — no consumer-customisable HTML anywhere —
-// so this CSP is unconditionally safe and gives the auth origin defense-in-depth
-// against any future trust-boundary regression that might emit a renderable body.
+// on every response. The go-pkgz/auth package's own response surface is JSON-only
+// for auth routes and images for the avatar route — no built-in HTML rendering
+// anywhere — so this CSP is unconditionally safe and gives the auth origin
+// defense-in-depth against any future trust-boundary regression that might emit a
+// renderable body.
 //
 //   - Content-Security-Policy: default-src 'none'; sandbox — blocks inline scripts
 //     and event handlers even if a body is ever served as HTML by mistake; the
@@ -254,6 +255,14 @@ func (s *Service) Handlers() (authHandler, avatarHandler http.Handler) {
 // The avatar Handler additionally sets Content-Disposition: inline; filename="avatar"
 // inside itself, so direct callers (tests, custom mounts) still get the full header
 // set without going through this wrapper.
+//
+// CONSUMER NOTE: custom providers added via Service.AddCustomHandler / AddProvider
+// are also wrapped. If a custom provider renders HTML (login forms, JS-based flows,
+// the dev_provider's login page, etc.), the strict CSP will block inline scripts and
+// event handlers on those pages. Such providers should either (a) override the CSP
+// for their own response by calling w.Header().Set("Content-Security-Policy", ...)
+// before writing — Set replaces the wrapper's value — or (b) move any required
+// scripts/styles to external files served from 'self'.
 func withSecurityHeaders(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Content-Security-Policy", "default-src 'none'; sandbox; frame-ancestors 'none'")
--- a/auth.go
+++ b/auth.go
@@ -236,7 +236,30 @@ func (s *Service) Handlers() (authHandler, avatarHandler http.Handler) {
 		p.Handler(w, r)
 	}
 
-	return http.HandlerFunc(ah), http.HandlerFunc(s.avatarProxy.Handler)
+	return withSecurityHeaders(http.HandlerFunc(ah)), withSecurityHeaders(http.HandlerFunc(s.avatarProxy.Handler))
+}
+
+// withSecurityHeaders wraps an auth response handler to apply strict CSP and nosniff
+// on every response. The go-pkgz/auth package response surface is JSON-only for auth
+// routes and images for the avatar route — no consumer-customisable HTML anywhere —
+// so this CSP is unconditionally safe and gives the auth origin defense-in-depth
+// against any future trust-boundary regression that might emit a renderable body.
+//
+//   - Content-Security-Policy: default-src 'none'; sandbox — blocks inline scripts
+//     and event handlers even if a body is ever served as HTML by mistake; the
+//     sandbox directive additionally isolates any rendered document from this origin.
+//   - X-Content-Type-Options: nosniff — prevents browsers from MIME-overriding the
+//     declared Content-Type to a more dangerous one.
+//
+// The avatar Handler additionally sets Content-Disposition: inline; filename="avatar"
+// inside itself, so direct callers (tests, custom mounts) still get the full header
+// set without going through this wrapper.
+func withSecurityHeaders(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Security-Policy", "default-src 'none'; sandbox; frame-ancestors 'none'")
+		w.Header().Set("X-Content-Type-Options", "nosniff")
+		next.ServeHTTP(w, r)
+	})
 }
 
 // Middleware returns auth middleware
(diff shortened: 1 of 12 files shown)


umputun (changes requested):
couple things to fix before merge:

1. Discord avatars look broken by this change. `provider/providers.go:289` builds `.webp` URLs, but `avatar/store.go` only registers GIF/JPEG/PNG decoders and `avatar/avatar.go:269` now requires `image.DecodeConfig` before storing. That means Discord WebP avatars will fall back to identicons even though `safeImgContentType` says WebP is allowed. Same in v2. This should be fixed in the avatar validation path, for example by registering `golang.org/x/image/webp` in both modules and adding a regression test.

2. `If-None-Match` compares against the unquoted raw store id (`avatar/avatar.go:224-230`, v2 same), while the response sends a quoted ETag. Standard clients send `If-None-Match: "<etag>"`, sometimes weak or comma-separated. This should parse validators or at least compare against the quoted value. Add tests for quoted and weak/multiple forms.

3. `safeImgContentType` says it is an allowlist of raster formats, but it accepts any `image/*` except SVG (`avatar/avatar.go:315-328`, v2 same). For this security fix I’d make it an explicit switch over PNG/JPEG/GIF/WebP/BMP/ICO.

4. Since global CSP is intentional, pls document the `AddCustomHandler` impact. `Service.Handlers()` now applies `default-src 'none'; sandbox` to custom provider handlers too, so HTML/form/script based custom flows can break.

minor: `var tinyPNG, _ = ...` in Telegram tests should use a `mustDecode` helper. Also consider using `io.ReadFull`/`ReadAtLeast` for the sniff buffer instead of a single `Read`, so a short read from a store doesn't reject a valid avatar.

umputun (approved):
lgtm, thx. Requested fixes look addressed.

PR #295, opened 2026-06-22
fix(example): incorrect callback urls after stripping route prefix

persidskiy:
Fixes the wrong redirect url in the example.

Step to reproduce:
1. set up google provider in `_example/main.go`, run the app
2. open http://localhost:8080/web, click google
3. google oauth handler errors with `Error 400: redirect_uri_mismatch`

This happens due to `r.URL.Path` being `/google/login` instead of `/auth/google/login` in `LoginHander` https://github.com/go-pkgz/auth/blob/master/v2/provider/oauth2.go#L143 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,8 +106,8 @@ func main() {
 
 	// setup auth routes
 	authRoutes, avaRoutes := service.Handlers()
-	router.Handle("/auth/", http.StripPrefix("/auth", authRoutes))   // add auth handlers
-	router.Handle("/avatar/", http.StripPrefix("/avatar", avaRoutes)) // add avatar handler
+	router.Handle("/auth/", authRoutes) // add auth handlers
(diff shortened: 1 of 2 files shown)


paskal (approved):
Verified — the auth handler dispatches on the last two path segments (`provName := elems[len-2]`), so dropping `StripPrefix` doesn't affect route matching; it fixes `makeRedirURL`, which builds the callback from the full `r.URL.Path`, to emit `/auth/{provider}/callback` instead of `/{provider}/callback` — the `redirect_uri_mismatch`. Example + README only, no runtime code. lgtm

umputun (approved):
lgtm. Confirmed both routes still dispatch on trailing path segments. `auth.go:231` reads the provider from `elems[len-2]` and `avatar.go:181` reads the id from the last segment, so dropping `StripPrefix` doesn't change routing, it only fixes the callback path built by `makeRedirURL`.

thx

PR #296, opened 2026-07-01
feat(avatar): add ErrNotFound sentinel returned by Store.Remove

paskal:
## What

The three avatar stores each reported a missing avatar differently from `Remove`:

- `LocalFS` — `os.Remove` → `os.ErrNotExist`
- `BoltDB` — `fmt.Errorf("avatar key not found, %s", ...)` (bare string)
- `GridFS` — `fmt.Errorf("avatar %s not found", ...)` (bare string)

So a caller had no reliable, portable way to tell "avatar already gone" from a genuine failure — it would have to string-match, and even that differs per store.

## Change

- Add an exported `ErrNotFound` sentinel.
- Return it (wrapped with `%w`) from all three `Remove` implementations when the avatar does not exist, so callers can use `errors.Is(err, avatar.ErrNotFound)`.
- Strengthen the existing `Remove` not-found tests from `assert.Error` to `assert.ErrorIs(..., ErrNotFound)`.

## Why

Downstream (e.g. remark42's account-deletion flow) needs to treat an already-removed avatar as success while still surfacing real errors. Today that's impossible to do portably across the stores; this makes it a one-liner.

### DIFF
--- a/v2/avatar/bolt.go
+++ b/v2/avatar/bolt.go
@@ -103,7 +103,7 @@ func (b *BoltDB) Remove(avatarID string) (err error) {
 	return b.db.Update(func(tx *bolt.Tx) error {
 		bkt := tx.Bucket([]byte(avatarsBktName))
 		if bkt.Get([]byte(avatarID)) == nil {
-			return fmt.Errorf("avatar key not found, %s", avatarID)
+			return fmt.Errorf("avatar %s not found: %w", avatarID, ErrNotFound)
 		}
 		if err = tx.Bucket([]byte(avatarsBktName)).Delete([]byte(avatarID)); err != nil {
 			return fmt.Errorf("can't delete avatar object %s: %w", avatarID, err)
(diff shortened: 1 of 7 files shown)


umputun (commented):
lgtm. One optional thing - `LocalFS.Remove` used to surface `os.ErrNotExist`, now the not-found path wraps only `ErrNotFound`, so anyone checking `errors.Is(err, os.ErrNotExist)` on the localfs store silently breaks. Keeping the original error in the chain is free with multi-%w:

```go
return fmt.Errorf("avatar %s not found: %w: %w", avatar, ErrNotFound, err)
```

not a blocker, fine as-is if you'd rather keep the message clean.

paskal:
Good point — applied (`26d85ca`). `LocalFS.Remove` now wraps both with multi-`%w` (`avatar %s not found: %w: %w`), so `errors.Is` matches `ErrNotFound` and the original `os.ErrNotExist`. Test asserts both. `-race`/vet/lint clean.

umputun (approved):
lgtm, thx

PR #300, opened 2026-07-10
ci: go 1.26, golangci-lint v2.12, address new gosec findings

paskal:
CI pinned go 1.24 while all three modules require 1.25 since #297, which made the example build fail on every dependabot PR until toolchain auto-download kicked in, and golangci-lint v2.6.2 panics on the go 1.26 stdlib.

Changes:

- CI: go 1.24 → 1.26, `golangci-lint` v2.6.2 → v2.12.2, `golangci-lint-action` v7 → v9 (v8/v9 have no breaking config changes, node runtime only), `--config=` form for args
- push triggers: explicit `branches: ["**"]` / `tags: ["**"]` instead of empty keys flagged by actionlint
- remaining actions to current majors: `checkout` v4 → v7, `codeql-action` v3 (deprecated) → v4; quoted `$GITHUB_WORKSPACE` in run scripts, actionlint now fully clean (`start-mongoDB@v0.2` is already latest)
- gosec findings new in v2.12, both v1 and v2 modules:
  - dev/test avatar endpoints (`dev_provider.go`, `custom_server.go`) now set `Content-Type: image/png` before writing the generated identicon
  - false positives suppressed with reasons: `token/jwt.go` cookies (Secure/SameSite come from service config, XSRF cookie is JS-readable by design), `provider/providers.go` oauth endpoint urls flagged as hardcoded credentials (G101), `provider/apple.go` redirect to the fixed apple auth endpoint (G710), `avatar/localfs.go` store-generated paths (G703)
  - gosec excluded in `_test.go` files via `.golangci.yml`
- gofmt on `avatar/avatar_test.go` (pre-existing comment alignment drift)

Both modules lint clean with v2.12.2, tests pass with `-race`, example builds. `go fix ./...` produced no changes.

### DIFF
--- a/.github/workflows/ci-v2.yml
+++ b/.github/workflows/ci-v2.yml
@@ -2,8 +2,8 @@ name: build-v2
 
 on:
   push:
-    branches:
-    tags:
+    branches: ["**"]
+    tags: ["**"]
     paths:
       - ".github/workflows/ci-v2.yml"
       - "v2/**"
@@ -59,7 +59,7 @@ jobs:
         uses: golangci/golangci-lint-action@v9
         with:
           version: v2.12.2
-          args: --config ../.golangci.yml
+          args: --config=../.golangci.yml
           working-directory: v2
 
       - name: submit coverage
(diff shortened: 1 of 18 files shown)


paskal:
@umputun, is it good to be merged?

umputun (approved):

